PR #2, opened 2024-12-01
Cicle wallpaper and theme

Davi-S:
# Pull Request

## Description

This pull aims to append two new scripts that can cycle through all installed wallpapers and themes and set them for the desktop and SDDM Corners theme. This is useful for those who want to see all wallpapers and themes they have installed or just is undecided on which one to use; for the latter, the user can set a cronjob to run the script periodically, thus having a new look on its PC from time to time.

(SDDM wallpaper will match the one on the desktop)

## Type of change

## Checklist

### DIFF
--- a/Configs/.local/lib/hyde/cyclewallpaperandtheme.sh
+++ b/Configs/.local/lib/hyde/cyclewallpaperandtheme.sh
@@ -1,86 +1,103 @@
 #!/usr/bin/env sh
 
-# Paths to "hyde" scripts
+# Paths to scripts and directories
 WALL_SCRIPT="$HOME/.local/share/bin/swwwallpaper.sh"
 THEME_SCRIPT="$HOME/.local/share/bin/themeswitch.sh"
-
-# Log file for debugging
+#SDDM_SCRIPT="$HOME/bin/setsddmwallpaper"
+SDDM_SCRIPT="$HOME/.local/share/bin/setsddmwallpaper.sh"
 LOG_FILE="$HOME/.local/share/cyclewallpaperandtheme.log"
-
-# State tracking files
 WALL_CYCLE_FILE="$HOME/.local/share/cyclewallpapercount"
 THEME_CYCLE_FILE="$HOME/.local/share/cyclethemecount"
 THEME_CHANGE_FLAG_FILE="$HOME/.local/share/cyclethemeflag"
-
-# Directory where themes are stored
 THEMES_DIR="$HOME/.config/hyde/themes"
+HYDE_CONFIG="$HOME/.config/hyde/hyde.conf"
 
-# If it's time to change the theme
-if [ -f "$THEME_CHANGE_FLAG_FILE" ]; then
-    # Get the list of themes and cycle through them (ensure spaces in file names are treated correctly)
+# Log messages to file
+log() {
+    echo "$(date '+%Y-%m-%d %H:%M:%S') $1" >> "$LOG_FILE"
+}
+
+# Safely read a counter file or initialize it
+read_or_initialize_counter() {
+    local file="$1"
+    [ ! -f "$file" ] && echo 0 > "$file"
+    cat "$file"
+}
+
+# Increment and save a counter
+increment_counter() {
+    local count="$1"
+    local max="$2"
+    local file="$3"
+    echo $(( (count + 1) % max )) > "$file"
+}
+
+# Cycle to the next theme
+cycle_theme() {
+    # Get theme list
     themes=()
     while IFS= read -r theme; do
         themes+=("$theme")
     done < <(find "$THEMES_DIR" -mindepth 1 -maxdepth 1 -type d | sort)
     themes=("${themes[@]##*/}")  # Strip path to get only names
 
-    # Read or initialize the theme counter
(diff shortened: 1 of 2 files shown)


kRHYME7:
Thank you for using snake_case! I need to rework some scripts to follow this convention. Never tested this but will merge this for now.

Configs/.local/lib/hyde/cyclewallpaperandtheme.sh
```diff
@@ -0,0 +1,103 @@
+#!/usr/bin/env sh
+
```
kRHYME7: Use bash in your shebang if the script is not POSIX (sh) at all, so the linters won't scream. 
```suggestion
#!/usr/bin/env bash

```

Configs/.local/lib/hyde/cyclewallpaperandtheme.sh
```diff
@@ -0,0 +1,103 @@
...
+WALL_SCRIPT="$HOME/.local/share/bin/swwwallpaper.sh"
+THEME_SCRIPT="$HOME/.local/share/bin/themeswitch.sh"
+#SDDM_SCRIPT="$HOME/bin/setsddmwallpaper"
+SDDM_SCRIPT="$HOME/.local/share/bin/setsddmwallpaper.sh"
+LOG_FILE="$HOME/.local/share/cyclewallpaperandtheme.log"
+WALL_CYCLE_FILE="$HOME/.local/share/cyclewallpapercount"
+THEME_CYCLE_FILE="$HOME/.local/share/cyclethemecount"
+THEME_CHANGE_FLAG_FILE="$HOME/.local/share/cyclethemeflag"
+THEMES_DIR="$HOME/.config/hyde/themes"
+HYDE_CONFIG="$HOME/.config/hyde/hyde.conf"
```
kRHYME7: Too hardcoded, I will expose a global variables later and will use it so global stuff can be source/imported. Also don't put logs in .local/share.

Configs/.local/lib/hyde/setsddmwallpaper.sh
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env sh
+
+# Configuration
+SDDM_THEME_CONFIG="/usr/share/sddm/themes/Corners/theme.conf"
+BACKGROUND_DIR="/usr/share/sddm/themes/Corners/backgrounds"
+WALLPAPER_NAME="customwallpaper"
+
```
kRHYME7: HyDE does have Candy theme too. how to handle it. This is the reason why there's 2 sddm wallpaper setter script in the PR. Both don't handle the other theme. Also this is the reason @Prasanth remove the script. We should rethink how to properly do it.

Anshul-007:
Can you switch to`#! /bin/env bash`, I think we are sticking to this convention.

kRHYME7:
@Davi-S  Please resolve reviews. I won't merge the sddm script as this is not a good approach, you can remove it for now so we can move forward

kRHYME7 (changes requested):

Configs/.local/lib/hyde/setsddmwallpaper.sh
Davi-S: Would it be better to make the "cyclewallpaperandtheme.sh" have a display manager script as parameter?
This way the user can create/edit the scripts as he likes to make them work for any theme or display manager he uses. AKA: Dependency Injection to achieve Inversion of Control

kRHYME7: @Davi-S  Hi! Can you separate the sddm wallpaper PR. So we can discuss on the wallpaper stuff.

Davi-S: should I just remove it or create a new pr?

kRHYME7: @Davi-S  Just remove it for now, I won't  merge the sddm yet. Will be building the CLI for that.  


---


> Would it be better to make the "cyclewallpaperandtheme.sh" have a display manager script as parameter?
This way the user can create/edit the scripts as he likes to make them work for any theme or display manager he uses. AKA: Dependency Injection to achieve Inversion of Control


About this one, No. the best approach IMO is to expose the wallpaper in ` /var/tmp/ `. and let sddm use `/var/tmp/hyde_wallpaper.png` 


#### Pros
-  no string attached between the user and sddm. 
- No need of root access to change stuff
- to implement this, we can do a one time change in the sddms theme's `theme.conf` file and set ` Background = /var/tmp/hyde_wallpaper.png ` 

##### Cons 
- Wallpaper will be exposed to everyone who had access to `/var/tmp`, but It's just a copy of the wallpaper so it should be fine, right?

Davi-S: The last time I checked, sddm could only view images in the theme folder; soft links didn't work either. It might be something with my system—I'll try it again.

kRHYME7:
As of the moment a while loop should cycle wallpapers. Not using `--global` flag will not change other colors only wallpaper which is helpful to avoid unexpected hitches.

PR #3, opened 2024-12-08
replace spotify package with spotify-launcher

JoeArchV:
##### 
241208_00h52m32s_screenshot
In the hyprdots the spotify package has a bug in the configuration section, the idea is to do this with a better user experience


kRHYME7:
This is not a fix, you're just doing a workaround by using xwayland for Spotify as spotify-launcher launches Spotify in xwayland by default.

PR #10, opened 2024-12-11
Animation management

kRHYME7:
# Pull Request
Based on @rishav12s'   https://github.com/prasanthrangan/hyprdots/pull/2043/files

https://github.com/user-attachments/assets/20964bd6-bbcc-44d4-906a-c52bf734b01c

if themes have $ANIMATION=
it will be used if present in ./hypr/animations


kRHYME7:
@rishav12s  Adding file.conf in the ~/.config/hypr/animations/*.conf

Now we just need to play the values of Variable named "$ANIMATION" in ` ./hypr/hyde.conf  ` 

it will be dynamically sourced upon reload but won't throw errors if files are not present.
IMO this is better than using whole "source=/file/path" string to do the job.

PR #19, opened 2024-12-13
Port waybar weather module pr from hyprdots

boulder-roller:
Port of prasanthrangan/hyprdots#1818
## Type of change


kRHYME7:
We should further look into this. 2 external dependencies are too much IMO.

Let's hold this PR for a while.

boulder-roller:
@kRHYME7 only one of them is actually used atm (python-requests), but in the original pr it was recommended to use the other, the script just hasn't been updated to include that yet.

kRHYME7:
@boulder-roller  Will merge this now. But there will be changes

Also, I think I should rename the "custom_hypr.lst : and "custom_app.lst ", it's confusing. 
cutsom_hypr.lst are the core apps and custom_apps.lst are optional.

PR #29, opened 2024-12-15
authorization from wayland

JoeArchV:
run scripts and applications that do not detect resolution, this is very good and recommended. I use auditing tools and that is necessary even for applications 
241215_13h09m21s_screenshot
xhost +SI:localuser:root
241215_13h12m04s_screenshot

### DIFF
--- a/Scripts/pkg_core.lst
+++ b/Scripts/pkg_core.lst
@@ -48,6 +48,7 @@ qt5-imageformats                                       # for dolphin image thumb
 ffmpegthumbs                                           # for dolphin video thumbnails
 kde-cli-tools                                          # for dolphin file type defaults
 libnotify                                              # for notifications
+xorg-xhost                                             # authorization from wayland
 
 # --------------------------------------------------- // Theming
 nwg-look                                               # gtk configuration tool


kRHYME7:
We have polkit in wayland, can you elaborate more for the certain use case? who benefits this? should we have this by default?

JoeArchV:
no brother, some programs in hyperland do not detect resolution through wayland and that is a solution

JoeArchV:
no brother, some programs in hyperland do not detect resolution through wayland and that is a solution

JoeArchV:
Programs of this type, when detecting that package, open it and start it without problems

JoeArchV:
not all 
1 fluxion 
2 airgeddon 
3gparted
Can I make Hyde start that after turning on the PC

JoeArchV:
I understood better if you say the apps x11, it is assumed that if you get the authorization from wayland

imtheabdi:
> Can you provide some of the programs? Are they all X11 apps?.

i use it to start Timeshift. it doesn't open otherwise. And if i remember correctly, BleachBit too, but i might be wrong here.

kRHYME7:
We can still add this in the pkg_custom.lst
 but should be  commented out.

Using this specifically to run those legacy apps (they won't fix it 😭) had some security risk,they say. Wayland protocol suck at this part. For usability of such packages, we can add this to the optional package list.

imtheabdi:
> We can still add this in the pkg_custom.lst
>  but should be  commented out.
> 

i agree.


> Using this specifically to run those legacy apps (they won't fix it 😭) had some security risk,they say. Wayland protocol suck at this part. For usability of such packages, we can add this to the optional package list.
> 
> We can't add this to the core, as obviously this is x11 workaround and not related to the dotfiles.
> 
> 

sure sure, that would be the better option. this isn't required for the proper functionality of the dots so it shouldn't be in the core apps.

PR #36, opened 2024-12-18
fix: Fix activation of sunset on startup

Anshul-007:
# Pull Request

## Description

- added code for launching hyprsunset on startup if user kept it running while he shutdown.
- previously it read the json file on startup and if user mode is set then it doesn't take action to enable it. So now we can read it from runnig instance of hyprsunset and compare it with json if not comaparable. Thus we can enable hyprsunset

## Type of change

## Checklist

## Screenshots

(if appropriate)


kRHYME7:
@Anshul-007  BTW why did you capped it to 10k ?  max is 20k

kRHYME7:
> Should i push it to 20k as well 

yes

kRHYME7:
ready for merging?

kRHYME7:
BTW I wanted to add something to this script later.

basically we need to use percent or 1-100 as values.

It's easy though, It's already implemented on the old hyprsunset.sh

this is more intuitive than learning how much 500 steps is.

Anshul-007:
Will look into it.

PR #46, opened 2024-12-21
fix monitor screenshot in screenshot.sh

rishav12s:
@kRHYME7 the monitor screenshots were not working only snip one worked

is this correct now ?


kRHYME7:
I intentionally avoid globbing/word splitting, and hope no one figures this out lol. Yes this is the fix for custom args to be passed.

PR #47, opened 2024-12-22
better suspend command 

rishav12s:
@kRHYME7 could you test it ?

i also have this in userprefs as i use a laptop maybe we could include this in keybindings.conf and comment it so if laptop users want a "sleep mode" more similar to that windows which first calls hyprlock and then suspends so whenever we wake the system it firsts asks for password

`bindl = , XF86PowerOff, exec, hyprlock & disown && systemctl suspend # launch lock screen & suspend`

### DIFF
--- a/Configs/.config/wlogout/layout_1
+++ b/Configs/.config/wlogout/layout_1
@@ -14,7 +14,7 @@
 
 {
     "label": "suspend",
-    "action": "systemctl suspend",
+    "action": "hyprlock & disown && systemctl suspend",
     "text": "Suspend",
     "keybind": "u"
 }
--- a/Configs/.config/wlogout/layout_1
+++ b/Configs/.config/wlogout/layout_1
@@ -14,7 +14,7 @@
 
 {
     "label": "suspend",
-    "action": "hyprlock & disown && systemctl suspend",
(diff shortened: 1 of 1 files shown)


kRHYME7:
use lockscreen.sh not hyprlock

rishav12s:
> use lockscreen.sh not hyprlock.
> 
> 
> Mine works fine. I got a locked session after suspend.
> 
> https://github.com/HyDE-Project/HyDE/blob/master/Configs%2F.config%2Fhypr%2Fhypridle.conf#L5-L11
> 
> But It's fine if we add it in wlogout,(example users don't have hypridle.

so like this 

`lockscreen.sh & disown && systemctl suspend`

for the keybinding i was talking about should we add it as a comment ?

or we could add a laptop.conf sometime later (having optimization targetted specially towards laptops)

 there should be a option for something like this in install.sh no ?

kRHYME7:
Yup we can or maybe add a docu for laptops.

imtheabdi:
> Yup we can or maybe add a docu for laptops.  

an install option asking users if they have a laptop then installing stuff like auto-cpufreq? if that doesn't make the install options too bloated.

kRHYME7:
Let's package the dotfiles into consumables. that way they can just pick what group of file they need.

PR #49, opened 2024-12-23
HyDE repository icon

JoeArchV:
With one click you will be sent to the HyDE repository
241222_23h51m30s_screenshot


imtheabdi:
i don't think it is a good idea for it to be in the Waybar. Besides, it looks ugly because it's not themed. And stuff like this is why browser bookmarks were made.

JoeArchV:
and when you just format it you save a few seconds less

JoeArchV:
The only thing I have in bookmarks is Netflix

JoeArchV:
It would be better if it was the wiki

kRHYME7:
1. The PR is wrong.
make a separate jsonc file in waybay/modules/

2. I don't mind adding the module as it can be use as proof of concept on making such custom modules. But It's debatable to add it by default in the waybar.

JoeArchV:
I will create the file and send pr

PR #65, opened 2024-12-28
Update hyde.conf

### LINKED ISSUE #63: [Bug]: Mouse
There is a bug that makes the mouse very big

JoeArchV:
cursor size fixed!

closes #63

### DIFF
--- a/Configs/.config/hypr/hyde.conf
+++ b/Configs/.config/hypr/hyde.conf
@@ -111,7 +111,7 @@
 # // █▄▄ █▄█ █▀▄ ▄█ █▄█ █▀▄
 
 #$CURSOR_THEME=Bibata-Modern-Ice
-#$CURSOR_SIZE=30
+#$CURSOR_SIZE=20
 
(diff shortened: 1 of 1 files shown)


kRHYME7 (approved):

kRHYME7:
try to use 24 is it the same as 20?

JoeArchV:
I will change it to 24

JoeArchV:
ready

kRHYME7:
Try to do it in your local.

Basically this is  not a fix. It only works with monitors with  scaling of 1. That's why the default is bigger as I'm using hidpi monitor. It is fone though assuming that most of the users have monitors with x1 scaling.

kRHYME7:
> ready


wrong file do it in this one

https://github.com/HyDE-Project/HyDE/blob/d6243beb5ea22d3132e9798644f1373d4ee816d3/Configs/.local/share/hyde/hyprland.conf#L115

JoeArchV:
The code view is read-only.

PR #72, opened 2024-12-31
breaking (zshrc): Rework for the zshrc implementation

kRHYME7:
breaks: zshrc should manually be cleaned up or else omz will reinitialize the plugins.   

for this compromise
+ I also added a warning if loading takes up more than 3 seconds for slow PC (works on a VM) 
+ Only showed 1 time per reboot 
image

feat: Won't touch .zshrc
feat: easy updates for hyde
feat: User configuration
feat: can uninstall hyde without removing users zshrc
feat: put oh-my-zsh locally ($HOME)
feat: update omz on restore_shl.sh
feat: faster startup by only using single `alias` call 

fix: omz packaging issues,

### DIFF
--- a/Configs/.hyde.zshrc
+++ b/Configs/.hyde.zshrc
@@ -1,0 +1,46 @@
+#  Startup 
+# Commands on startup (before the prompt is shown)
+# This is a good place to load graphic/ascii art, display system information, etc.
+
+# pokego --no-title -r 1,3,6
+# fastfetch --logo-type kitty
+# fastfetch.sh
+
+#  Aliases 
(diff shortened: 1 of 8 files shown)


kRHYME7:
@Anshul-007  @T-Crypt  @rubiin  @rishav12s  @abenezerw 

🙏

kRHYME7:
Tested on a VM with ` install.sh -r `    


Note that this .zshrc should be cleaned.

rubiin (approved):

kRHYME7:
merging this now, please report issues. Thanks!

PR #79, opened 2025-01-02
Update window.jsonc

JoeArchV:
netflix, discord and chrome added

### DIFF
--- a/Configs/.config/waybar/modules/window.jsonc
+++ b/Configs/.config/waybar/modules/window.jsonc
@@ -15,6 +15,8 @@
             "(.*)Netflix": "Netflix 󰝆 ",
             "(.*) — Google chrome": "$1  ",
             "(.*)Google chrome": "Google chrome  ",
+            "(.*) — Github": "$1  ",
+            "(.*)Github": "Github ", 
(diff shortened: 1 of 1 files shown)


kRHYME7 (approved):
Thanks!

Anshul-007:
@kRHYME7 if the restore_config.psv has a sync flag enabled this wouldn't be getting overwritten right?

kRHYME7:
This will be overwritten.

for now same as hyprdots I added an option to make a userpref for modules

example


`modules/windows.jsonc ` -> use by hyde

users can cop above module and modify.

`modules/windows.jsonc#my_custom_module` -> Hyde don't know about this tag so it won't be overwritten.


in config.ctl


```...windows#my_custom_module ...```


this will now use your custom copy of the module

sitolam:
> @kRHYME7 if the restore_config.psv has a sync flag enabled this wouldn't be getting overwritten right?
> 

What does the .psv mean?

kRHYME7:
@sitolam open the file.

It should still .lst but to not break other stuff just renamed it to .psv.

hint: read the file

PR #81, opened 2025-01-03
Update install.sh

JoeArchV:
avoid future problems


JoeArchV:
What do you think about this? It would be good for new users.

Anshul-007:
can do but instead of doing nothing if user cancels we can do `hyde-shell -r`

JoeArchV:
Yes, if the user cancels, nothing happens, it is just a recommendation.

kRHYME7 (commented):
use english variable name

JoeArchV:
change respuesta?

JoeArchV:
listo

JoeArchV:
list

JoeArchV:
read answer

if [ "$answer" == "y" ]; then

kRHYME7 (approved):
thankss

PR #86, opened 2025-01-05
ocapacity

JoeArchV:
0.85 0.75

250102_14h18m28s_screenshot

### DIFF
--- a/Configs/.local/share/hyde/hyprland.conf
+++ b/Configs/.local/share/hyde/hyprland.conf
@@ -145,8 +145,8 @@ $FONT_HINTING=$default.FONT_HINTING
 
 decoration {
     dim_special = 0.3
-    active_opacity = 0.95
-    inactive_opacity = 0.80
+    active_opacity = 0.85
+    inactive_opacity = 0.75
     fullscreen_opacity = 1
     blur {
         special = true


JoeArchV:
250102_14h18m28s_screenshot

JoeArchV:
The current configuration does not look good, it is as if it did not have

kRHYME7:
I defaulted the opacity for the active window as 0.95 for usability. Also don't change values here as themes override this. 

So opacity on Windows is controlled by themes.

kRHYME7:
If you want to use those values add them in userprefs.conf for user preference.

JoeArchV:
I'm here to make it look better

JoeArchV:
If that is already the preference of the users

kRHYME7:
This is theme specific so changing this here is useless. This is only for fallback if nothing is set. so maybe try to use another theme

kRHYME7 (approved):
x2 scaling monitors will suffer with this change but looks good on x1 scaling. So I'll still merge


BTW, I'm on an x2 scaling monitor so the opacity value is multiplied too. Therefore visuals for x2 scaling unusable sometimes. But assuming most monitors are x1 so I think it is fine. 

For future reference for other users whom  scaling  != 1
add this to your `.config/hypr/hyde.conf`

```
decoration {
active_opacity = 0.95
inactive_opacity = 0.80 


}
```


adjust the values that are suitable for you.

kRHYME7:
This might be a bad idea,

PR #93, opened 2025-01-08
neofetch config

JoeArchV:
250102_14h18m28s_screenshot

### DIFF
--- a/Configs/.config/neofetch/config.conf
+++ b/Configs/.config/neofetch/config.conf
@@ -1,0 +1,115 @@
+print_info() {
+    prin "┌─────────\n Hardware Information \n─────────┐"
+    info " ​ ​ 󰇺 " model
+    info " ​ ​  " cpu
+    info " ​ ​ 󰊴 " gpu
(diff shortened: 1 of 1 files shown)


JoeArchV:
almost all fastfetch icons are in neofetch

kRHYME7:
because we are using neofetch before switching to fastfetch. We stop using neofetch because it is unmaintained.

Fun fact, the configs we have for fastfetch are translated from neofetch.

JoeArchV:
I use both

JoeArchV:
I know it's not maintained but I know there are people who use neofetch

JoeArchV:
One of those is me, I even use pfetch

kRHYME7:
if you PERSONALLY still want to use it, heres the config before moving to fastfetch.

https://github.com/prasanthrangan/hyprdots/pull/1343#pullrequestreview-2013059929


I won't be adding neofetch back unless there's a good reason for adding it back. For now fastfetch is performant and looks the same sa neofetch.

PR #100, opened 2025-01-12
Update systemupdate.sh

JoeArchV:
snap packet support


kRHYME7:
Hi won't be merging more stuff in systemupdate yet. I'm moving some stuf related to package managers so these  changes might be overwritten soon

kRHYME7:
BTW if you want to actually automate updating use `topgrade` I won't recommend it to be added here though, But I think you'll like it

JoeArchV:
Ok, so I can improve the code a little bit.

JoeArchV:
No, not just the support

Issue #108, opened 2025-01-16
[Improvement] CPU Info: Detect if AMD and make temperature display TccdX and not Tctl


Olaipai:
### Description

Make changes to the **cpuinfo.sh** script to detect if the processor is an AMD Ryzen (Zen) processor and make it use tCCDX (1 or 2).

### Use Case

tCTL on Ryzen is a combination of the current die with an added offset. The main purpose of this is to ensure that the CPU fan(s) are running as desired and making sure that fans are in use (tDIE + X (offset)). The offset is different from architectures.  
Zen1 offset = +20C
TR1 Offset = +27C
Zen3 offset = +15C (I think, this is the case on my 5900x at least)

As you can see, the offset is quite big. This means that if a AMD Ryzen user isn't aware of this, they might think that their CPU is running a lot hotter than what it in reality actually does. 

Some AMD CPUs have a multi chiplet design with multiple CCDs and therefore multiple tCCD(x). This is usually the case with the x9x0 series (5900x, 5950x,7900x,7950x etc...).
An exception is-  There exists a couple of 5600x processor which do have multiple CCDs, however they are not very common.

![image](https://github.com/user-attachments/assets/b008dec2-2ab5-423b-bd97-c1e3fbc15931)

### Proposed Solution

I am the epitome of dog shit at scripting and coding. However, I solved this by adding two new temp variables and printing them.

<details>
<summary>My temp solution</summary>

```
# CPU temp
sensorsdata=$(sensors)
temp1=$(awk '/Package id 0/ {print $4}' <<< "$sensorsdata" | awk -F '[+.]' '{print $2}')
temp2=$(awk '/Package id 0/ {print $4}' <<< "$sensorsdata" | awk -F '[+.]' '{print $2}')
if [ -z "$temp1" ]; then
    temp1=$(awk '/Tccd1/ {print $2}' <<< "$sensorsdata" | tr -d '+°C')
fi
if [ -z "$temp2" ]; then
    temp2=$(awk '/Tccd2/ {print $2}' <<< "$sensorsdata" | tr -d '+°C')
fi
if [ -z "$temp1" ]; then
    temp="N/A"
else
    temp=`printf "%.0f\n" $temp`
fi

[...]

PR #110, opened 2025-01-17
perf: Sensors Info module 

kRHYME7:
# Pull Request

POC for now and let me know what features you want to add

ref:
https://github.com/HyDE-Project/HyDE/issues/108

#### My sensors
image
image

#### Test from https://github.com/HyDE-Project/HyDE/issues/108#issuecomment-2595746955
image
image
image

## Benchmarks


Olaipai:
Ngl, these look great! I also dig the colors matching the temps. 
For more easy overview, it might be wise to hide empty values - `Fan speed: 0 rpm`, I like the implementation of multiple pages. 
Maybe once you get this working, one can implement toggles with mouse buttons, say "show all" and "show core (cpu, gpu, fans))

One idea once starting to rollout could be to use hyde-ctl to initially probe lm-sensors and setup a possible config of what the user wants to show. Almost like a promp similar to some guided prompt-configurators (Tide promp is a good example)

I can probably continue to brainstorm ideas. But one step at a time! ^_^

Not that this is important, but: The kernel module for aquacomputer stuff is partly working. Conductivity, Water quality and Flow are all divideable by 10 and none of them are measured in RPM as you can tell.

kRHYME7:
> One idea once starting to rollout could be to use hyde-ctl to initially probe lm-sensors and setup a possible config of what the user wants to show. Almost like a promp similar to some guided prompt-configurators (Tide promp is a good example)


That will be easy later. For now before I merge this maybe we can add in hyde.toml

```toml
[sensorsinfo]
chips=[
"chip1",
"chip2"
]
```


chipX are the exact device name.

kRHYME7:
That's why I added this one, as cpuinfo is more on generalized cpu temp stuff.

PR #113, opened 2025-01-18
Update pkg_core.lst

JoeArchV:
unrar


kRHYME7:
what script or program requires this?

JoeArchV:
I deleted Dolphin and it won't let me extract unrar files. I reinstalled it and it lets me.

kRHYME7:
For now there's no rar archive in the repo.Please don't bloat the repo with unnecessary stuff. Instead we should be trying to always watch what dependencies we really need.

JoeArchV:
It is not to remove dolphin unrar what dolphin requires to extract that type of files

PR #119, opened 2025-01-19
Update hyprlock.conf

JoeArchV:
Visualization for cava


JoeArchV:
I can't take a photo of hyprlock since I don't have a cell phone. I'll soon receive a Phone 11 from the United States.

kRHYME7:
Make a stand alone hyprlock layout and call the cava script in there. 

> [!CAUTION]
> $CAVA_CMD is bugyy as it is only a workaround

get some inspiration from here 

image

PR #130, opened 2025-01-21
Create hackthebox.jsonc

JoeArchV:
This module will detect whether or not you are connected to hack the box and tell you


JoeArchV:
I thought this for people who use hackthebox and for myself

kRHYME7:
> I thought this for people who use hackthebox and for myself 

We can add a different approach, but not this one. Majority in HyDE is not in networking, and most likely not using HTB.  This is too specific. 

I am into networking though, And I think a module to check if you're connected to a certain server will be cool. 
I'll probably implement similar to this one, but not this.

PR #131, opened 2025-01-21
Create hackthebox.sh

JoeArchV:
This is the script for the module


kRHYME7:
Note never hardcode dynamic device/interface either physical or virtual devices. tun0 is not always set by default in all systems.

JoeArchV:
That depends on the users, they can add the module whenever they want.

PR #136, opened 2025-01-22
Update pkg_extra.lst

JoeArchV:
spotify ad blocker


kRHYME7:
https://github.com/HyDE-Project/HyDE/discussions/137

PR #142, opened 2025-01-23
Used weber fechner law for better human perception brightness

Meshwa428:
used very littleish weber fechner law to help the brightness control be more appealing to human perception.

previos version required static values, this one calculates them on its own

### DIFF
--- a/Configs/.local/lib/hyde/brightnesscontrol.sh
+++ b/Configs/.local/lib/hyde/brightnesscontrol.sh
@@ -1,76 +1,88 @@
-#!/usr/bin/env bash
+#!/usr/bin/env sh
 
-scrDir=$(dirname "$(realpath "$0")")
-# shellcheck disable=SC1091
-source "$scrDir/globalcontrol.sh"
+# Check if the script is already running
+pgrep -cf "${0##*/}" | grep -qv 1 && echo "An instance of the script is already running..." && exit 1
+
+scrDir=`dirname "$(realpath "$0")"`
+source $scrDir/globalcontrol.sh
 
 # Check if SwayOSD is installed
 use_swayosd=false
-isNotify=${BRIGHTNESS_NOTIFY:-true}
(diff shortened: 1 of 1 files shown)


kRHYME7:
@Meshwa428, the current implementation has a fundamental misunderstanding of perception compensation.

The script is currently implementing the perceived brightness curve, which is incorrect. Instead, we need to compensate for the non-linear perception.

-    We don't want to replicate how humans perceive brightness
- We want to create a control mechanism that mathematically compensates for non-linear perception


What we should do: 
-    Counteract the logarithmic nature of human brightness perception
-    Create a more linear, predictable brightness adjustment mechanism

Implementation: 
Rather than following the perceived curve, we should:

-    Implement exponential or inverse logarithmic adjustments
-    Ensure more consistent perceived brightness changes
-    Provide uniform step sizes that FEEL more predictable to users

Example Principle:
When humans perceive small changes at low intensities, our script should make larger intensity changes to create a consistent perceived adjustment.

# The perception ( law )
image

# Expected Real Feel
This is how the user feels. It should feel like the script is actually changing values even on higher brightness 
image

# Technical part
We should design the script to handle the perception and  compensate that so the real feel or actual perception is linear and feels natural and effective

Meshwa428:
So we need something similar to what windows does?

Single press reduces 1 or 2 brightness points, and if consecutive presses then do perception thingy

kRHYME7:
> So we need something similar to what windows does?

Kind of

> Single press reduces 1 or 2 brightness points, and if consecutive presses or the button kept pressed down then do perception thingy

Almost, but not really. The issue here is we don't have any service/daemon that is already running and reliable to handle events like key presses. 

Therefore, what I proposed is to inverse log to compensate the log perception. Sort of making the script exponential on higher intensities. 

Sorry for nerding out XD, but just like in amps, some knobs are logarithmic potentiometers so that the users will perceive the change even on high amplitudes/volumes.   
image

kRHYME7:
Let's reiterate this idea first, as this is not really that of importance.  No need to complicate such simple tasks.

PR #143, opened 2025-01-23
Update backlight.jsonc

JoeArchV:
"on-click": "nwg-displays",


kRHYME7:
This is too unrelated to the backlight, make a separate module instead

JoeArchV:
ok

PR #162, opened 2025-01-26
warning handling python scripts

0xGeN02:
# Pull Request

## Description

-  Missing timeout argument for method 'requests.get' can cause your program to hang indefinitely

Solution adding a timeout of 10s to finish process in case of unfetching data

- 'subprocess.run' used without explicitly defining the value for 'check'.

When using subprocess.run, it's a good practice to explicitly define the check parameter to ensure that the function behaves as expected. The check parameter, when set to True, will raise a CalledProcessError if the command exits with a non-zero status.

- Use lazy % formatting in logging functions

This way, the actual string formatting is deferred until it is necessary, which can improve performance.

- E1310:bad-str-strip-call

The error E1310:bad-str-strip-call indicates that the argument passed to str.strip is suspicious. The strip method is used to remove leading and trailing characters from a string, and it takes a string of characters to be removed. In your code, you are passing " + " to strip, which is not a valid argument for this method.

To fix this, you should use str.replace instead of str.strip to remove the unwanted characters. Here is the corrected line:

`" + ".join(keys).replace(" + ", "").strip()`

This will replace occurrences of " + " with an empty string and then strip any leading or trailing whitespace.

WARN: it only works with python newer than 3.9

### DIFF
--- a/Configs/.local/lib/hyde/amdgpu.py
+++ b/Configs/.local/lib/hyde/amdgpu.py
@@ -75,9 +75,19 @@ def main():
 
         # Print the JSON string
         print(json_output)
-        
-    except Exception as e:
-        print(f"Error: {str(e)}")
     
+    except json.JSONDecodeError as e:  # Handle JSON decoding errors (e.g., invalid JSON)
+        print(f"JSON Error: {str(e)}")
+    except AttributeError as e:  # Handle attribute errors (e.g., method not found)
+        print(f"Attribute Error: {str(e)}")
+    except ValueError as e:  # Handle value errors (e.g., invalid value for formatting)
+        print(f"Value Error: {str(e)}")
+    except RuntimeError as e:  # Handle runtime errors (e.g., issues with querying the GPU)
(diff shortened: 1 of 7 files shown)


kRHYME7:
@Senshi111  will this break in debian ?

I checked in distro box and debians python version is 3.11.2,
image

Senshi111:
> @Senshi111 will this break in debian ?
> 
> I checked in distro box and debians python version is 3.11.2, image

I dont think so. debian experimental branch is in paython3 3.12.1. Dont worry about it. I still dont have debian set up yet. I have some work this weekend

kRHYME7:
@Senshi111  @0xGeN02  Will merge this then, let's just fix anything that might break.

Senshi111:
> @Senshi111 @0xGeN02 Will merge this then, let's just fix anything that might break.

ok. i already did for my fedora-hyde

0xGeN02:
If you find any issues related to this just tell me and I will try to fix them @kRHYME7 @Senshi111

PR #168, opened 2025-01-27
consistent shabangs

kRHYME7:
# Consistency 

- This is to support modern distro. 
- We need to have a common interpreter to debug correctly

####  For legacy distro
I think we can just manually link  `/bin/env`  to `/usr/bin/env`  or vice versa, just like how other distros do it. 

So my proposal will be in `install.sh` we will add a strict check if `/usr/bin/env` exist if not find it if error abort. 

But on debian12 it is /usr/bin/env is already supported. (using distrobox)
image
 
#### why use bash instead of sh ? 

Because most of the scripts that use `sh` are actually written in  `bash`! That means there will be unpredictable stuff. And debugging is annoying. 

I would like to discuss it here @Senshi111  @rubiin @richen604 

ref: https://github.com/HyDE-Project/HyDE/discussions/149#discussioncomment-11949678

### DIFF
--- a/Configs/.local/bin/hyde-shell
+++ b/Configs/.local/bin/hyde-shell
@@ -1,4 +1,4 @@
-#!/bin/env bash
+#!/usr/bin/env bash
 # This script will soon be use to resolve $PATHS and LIB_DIR specific to hyde
 # wallbash script can source this script to resolve the paths cleanly on a separate shell
 
--- a/Configs/.local/lib/hyde/cava.sh
+++ b/Configs/.local/lib/hyde/cava.sh
@@ -1,4 +1,4 @@
-#!/bin/env bash
+#!/usr/bin/env bash
 #----- Optimized bars animation without much CPU usage increase --------
 #----- Optimized bars animation without much CPU usage increase pt2 --------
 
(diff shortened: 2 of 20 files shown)


Senshi111:
My experience with sh on debian is that those scripts will not work. Never did for me when i tried.

Until i configure debian for testing hyde i can't say too much at the moment. I have lot of work to do this week, so i will not be so active.

kRHYME7:
> Until i configure debian for testing hyde i can't say too much at the moment. I have lot of work to do this week, so i will not be so active.

Appreciated the quick response! No pressure at all, take your time. Maybe I can try to spin up a VM later. Thank you, @Senshi111

JoeArchV:
I could try since I don't have classes this week

JoeArchV:
I will download debian later

richen604:
the abort is good idea, let the user know of the standard and they can decide if symlinking is the correct option for them

kRHYME7:
image

I think it's fine in debian12.

richen604:
that checks out with https://wiki.debian.org/UsrMerge
we probably won't see many systems using `/bin/env`, unless HyDE on BSD 😂

kRHYME7:
> unless HyDE on BSD 😂

unless, 🤔 Lol


Please add review so I can merge this one

kRHYME7:
merging it now ,

PR #208, opened 2025-02-04
Load OMZ in .hyde.zshrc

AhJi26:
I notice that after update Hyde, my custom sctipts in omz wouldn't load properly.

So I bring back the OMZ loading script at the top of .hyde.zshrc
I think this should be managed by hyde if we use OMZ by defualt.

### DIFF
--- a/Configs/.hyde.zshrc
+++ b/Configs/.hyde.zshrc
@@ -1,6 +1,3 @@
-#  Load OMZ Scripts 
-source $ZSH/oh-my-zsh.sh
-
 #  Startup 
 # Commands on startup (before the prompt is shown)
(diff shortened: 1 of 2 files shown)


kRHYME7:
ops not the correct way to do it. OMZ should be in .zshenv.  Can you point out which commit has changed that

kRHYME7:
Lemme know if ready to merge

kRHYME7:
do omz handle aliases? It is intended to put aliases somewhere the user can overwrite later

kRHYME7:
basically move alias on top of the OMZ loading? It's fine.

kRHYME7:
lemme knows if we can merge this now

kRHYME7:
Thank you!

PR #209, opened 2025-02-04
Fix dunst font setting invalid bug

AhJi26:
When modify the dunst font in donst.conf, the modified font will be added to dunstrc, but forced reset to defualt value.

Delete unnecessary font config and duplicate setting of dmenu(witch is already set at the top of the script).

### DIFF
--- a/Configs/.hyde.zshrc
+++ b/Configs/.hyde.zshrc
@@ -1,6 +1,3 @@
-#  Load OMZ Scripts 
-source $ZSH/oh-my-zsh.sh
-
 #  Startup 
 # Commands on startup (before the prompt is shown)
 # This is a good place to load graphic/ascii art, display system information, etc.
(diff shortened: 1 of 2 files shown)


kRHYME7:
You want static font or use font according to theme?

kRHYME7:
default font size of dunst is not working?

kRHYME7:
Will try to see if dunst supports a separate font scaling declaration if not we will use the FONT_SIZE declared in hyprland.

kRHYME7:
done

kRHYME7:
@AhJi26 

In your . config/hyde/config.toml


```toml
[notification]
font_size = 12

```

kRHYME7:
You're on latest commits?

kRHYME7:
try invoking this 
```
parse.config.py
```
If it exists, are you sure you edited `~/.config/hyde/config.toml` correctly?

kRHYME7:
My dunst is working correctly.  Let's wait a bit maybe someone can reproduce this also.

PR #211, opened 2025-02-05
Translate keybindings.md to Chinese

AhJi26:
Ready to merge.


kRHYME7:
BTW you can generate a table using `keybinds.hint.py --format md`.

kRHYME7:
Planning to add a different set of keybindings.conf so users can use it specially on english users. Or might someone have a different keyboard layout and keybindings.conf that just makes sense on that layout. That tool can be used later.

kRHYME7:
Nope, Binds should be treated differently. We can just add it in the repo and just add a part in the readme that there are pre built binds available and to choose from.

PR #223, opened 2025-02-07
Feat: Add config option for system monitor

AhJi26:
Add support for user to select a systom monitor with config option in ~/.config/hyde/config.toml like this.
```toml
[utils]
system_monitor = "btop"
```

### DIFF
--- a/Configs/.local/lib/hyde/gnome-terminal
+++ b/Configs/.local/lib/hyde/gnome-terminal
@@ -1,0 +1,22 @@
+#!/usr/bin/env bash
+# This is a wrapper script to be use by gtk-launch and other gtk stuff
+# that is expecting users to ALWAYS use the "gnome-terminal" command.
+# -- The HyDE Project --
+
+scrDir="$(dirname "$(realpath "$0")")"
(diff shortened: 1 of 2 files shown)


kRHYME7:
@AhJi26  This don't work? 

```
[sysmonitor]
#  sysmonlaunch.sh configuration // System monitor configuration
execute = "btop"                   # Default command to execute // accepts executable or app.desktop
commands = ["btop", "htop", "top"] # Fallback command options

```

https://github.com/HyDE-Project/HyDE/blob/9e05f58eb300c4907950ee29ff661d5f24f4cd89/Configs/.local/share/hyde/config.toml#L176C1-L180C1

kRHYME7:
Still broken haha, 


that's originally an array so no need to  make it a string then convert back to an array.

For example we have a priority variable `$X` and you want to add it as the first item.


```

# initiates an array consisting of a,b,c,
array=( "a" "b" "c" ) 

# this part we appended array2 to array
array+=( "${array2}" ) 

# Now, let's try to add X, and making it the priority. assuming X is not empty.
array+=( "$X" "${array}"  )



```

PR #239, opened 2025-02-11
Update keybindings.conf

mahaveergurjar:
Update keybinding to easy access

### DIFF
--- a/Configs/.config/hypr/keybindings.conf
+++ b/Configs/.config/hypr/keybindings.conf
@@ -27,14 +27,17 @@
 $wm=Window Management
 $d=[$wm]
 bindd = $mainMod, Q, $d close focused window, exec, $scrPath/dontkillsteam.sh
-bindd = Alt, F4, $d close focused window, exec, $scrPath/dontkillsteam.sh
+# bindd = Alt, F4, $d close focused window, exec, $scrPath/dontkillsteam.sh
 bindd = $mainMod, Delete, $d kill hyprland session, exit
 bindd = $mainMod, W, $d toggle float, togglefloating, #
 bindd = $mainMod, G, $d toggle group, togglegroup
-bindd = Shift, F11, $d toggle fullscreen, fullscreen
+# bindd = Shift, F11, $d toggle fullscreen, fullscreen
+bindd = ALT, return, $d toggle fullscreen, fullscreen
+bindd = $mainMod, return, $d toggle false fullscreen, fullscreen, 1
 bindd = $mainMod, L, $d lock screen, exec, lockscreen.sh
 bindd = $mainMod Shift, F, $d toggle pin on focused window, exec, $scrPath/windowpin.sh
 bindd = Control Alt, Delete, $d logout menu, exec, $scrPath/logoutlaunch.sh
+bindd = $mainMod, SPACE, $d logout menu, exec, $scrPath/logoutlaunch.sh 2
 bindd = Alt_R, Control_R, $d toggle waybar and reload config, exec, killall waybar || (env reload_flag=1 $scrPath/wbarconfgen.sh)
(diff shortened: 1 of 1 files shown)


Configs/.config/hypr/keybindings.conf
```diff
@@ -31,10 +31,12 @@ bindd = Alt, F4, $d close focused window, exec, $scrPath/dontkillsteam.sh
 bindd = $mainMod, Delete, $d kill hyprland session, exit
 bindd = $mainMod, W, $d toggle float, togglefloating, #
 bindd = $mainMod, G, $d toggle group, togglegroup
-bindd = Shift, F11, $d toggle fullscreen, fullscreen
```
kRHYME7, line 34: use this instead please 

```
$cyle_fullscreen=hyprctl activewindow -j | jq -r '.fullscreen | tonumber + 1 | if . > 2 then 0 else . end | tostring' | xargs -I{} hyprctl dispatch fullscreenstate {}
bindd = SUPER, F11, $d cycle fullscreen,exec, $cyle_fullscreen
```

Let's use this  to cycle fullscreenstate.  I don't like using all the keyboard keys, It's not efficient at all and too confusing. 

I prefered `SUPER +  F11 `  to cycle fullscreen. But feel free to propose something better. I would prefer something that is from windows or macos. Newbies should use something familiar, intermidiate users can just edit it.

Configs/.config/hypr/keybindings.conf
```diff
@@ -31,10 +31,12 @@ bindd = Alt, F4, $d close focused window, exec, $scrPath/dontkillsteam.sh
...
+bindd = ALT, return, $d toggle fullscreen, fullscreen
+bindd = $mainMod, return, $d toggle false fullscreen, fullscreen, 1
 bindd = $mainMod, L, $d lock screen, exec, lockscreen.sh
 bindd = $mainMod Shift, F, $d toggle pin on focused window, exec, $scrPath/windowpin.sh
 bindd = Control Alt, Delete, $d logout menu, exec, $scrPath/logoutlaunch.sh
+bindd = $mainMod, SPACE, $d logout menu, exec, $scrPath/logoutlaunch.sh 2
```
kRHYME7, line 39: we don't need a bind for everything by default. 

```toml
[wlogout]
style = 2 
```

this is used so that logoutlaunch will use style 2

kRHYME7 (changes requested):

Configs/.config/hypr/keybindings.conf
mahaveergurjar: ok

Configs/.config/hypr/keybindings.conf
mahaveergurjar: ok but i think replace Control Alt, Delete, with $mainMod, SPACE its more easy ???

kRHYME7: any reason  ` super + space `  is "easier"? 

I added `ctrl + alt + delete ` as a logout menu from windows. You see most of the users are from windows, I want an easier entry point for them to enjoy this dots.

mahaveergurjar: yeah due to two keys binding so user easily remember it

whoisabdi: i use alt + escape because I'm used to spam pressing alt + f4 to close everything and shutdown PC. since i can't use alt + f4 again for logout menu, i used the next key with alt that would be the easiest to reach and it's the escape key. i use escape because I'll have to look for other keys but with escape i can be sure that the left most key is escape. i use the same for fullscreening windows, windows + escape.

mahaveergurjar: Okay

PR #305, opened 2025-02-26
Minor fixes 

0xGeN02:
- Improved wttr.in data fetchin + name parisng to snake_case
- subprocess.run check handling
- imporve performance in array iterations
- fixed an unassigned use of a variable
- add utf-8 encoding as default

250226_00h50m03s_screenshot

### DIFF
--- a/Configs/.local/lib/hyde/sensorsinfo.py
+++ b/Configs/.local/lib/hyde/sensorsinfo.py
@@ -95,7 +95,7 @@ def get_temp_color(temp):
     return f"{temp}°C"
 
 
-def get_sensor_data(page=0):
+def get_sensor_data(result_sensors, page=0):
     try:
         sensors_data = json.loads(result_sensors.stdout)
     except json.JSONDecodeError:
@@ -222,5 +222,5 @@ if __name__ == "__main__":
         page = (page - 1 + total_pages) % total_pages
         subprocess.run(["pkill", "-RTMIN+19", "waybar"], check=False)
     save_current_page(page)
-    sensor_info = get_sensor_data(page)
+    sensor_info = get_sensor_data(result_sensors, page)
     print(json.dumps(sensor_info, separators=(",", ":")))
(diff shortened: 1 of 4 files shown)


Configs/.local/lib/hyde/sensorsinfo.py
kRHYME7, line 1: Please retain or improve the old performance. This is 2x slower. Note that sensors.py or any monitoring should at least be faster

image

kRHYME7 (changes requested):

Configs/.local/lib/hyde/weather.py
```diff
@@ -202,25 +203,33 @@ def format_chances(hour):
         os.getenv("WEATHER_FORECAST_DAYS", "3")
     )  # Number of days to show the forecast for (default: 3)
 except ValueError:
-    forecast_days = 3
```
kRHYME7, line 205: What is the scope of these variables. Iirc they are not global, right? Why CAPS?

kRHYME7 (changes requested):

Configs/.local/lib/hyde/sensorsinfo.py
0xGeN02: Okey will try with a workarround to improve performance

Configs/.local/lib/hyde/weather.py
0xGeN02: In python, all constants are supposed to be named as UPPER_CASE for convention (see https://peps.python.org/pep-0008/#constants)

Configs/.local/lib/hyde/sensorsinfo.py
kRHYME7: I already fixed it.

Configs/.local/lib/hyde/weather.py
kRHYME7: > Constants are usually defined on a module level and written in all capital letters with underscores separating words. 


But I think this is local in the function and not on a module level? no?.

kRHYME7: Ah yeah I see, in this context it is used globally. Is this ready for merge ?

0xGeN02: Yes it is used globally, all good for merge

PR #325, opened 2025-03-02
FORMAT and FIXES in JSON/JSONC

0xGeN02:
MAIN CHANGES

     - Formatting JSON and JSONC files to standard format
     - Fixing error in JSON fles: unexpected ${var} to match JSON type string "${var}"
     - Fixed unexpected types on fastfetch config files, I tried to not change the default styling config done by authors
     
     
250302_22h01m44s_screenshot

Before and after styles are still the same 

250302_21h57m25s_screenshot

Same here


kRHYME7:
For the waybar it will break rotation. I already tested this hence the rewrite of waybar.

0xGeN02:
How are you testing the rotation? As I see it is used in the `wbarconfgen.sh` but i don`t know how to test it properly, will try a workaround to make this function as expected

kRHYME7:
There's no workaround, wbarconfgen parses this as a simple text file not json/c. That's the reason for the rewrite.

PR #215, opened 2025-02-06
Waybar Rework

kRHYME7:
feat:
- Style selection
- Layout Selection
- Can override layout by using ~/.config/waybar/config
- Modify waybar manually
-  have extra groups styling, this is the pill like groupings/islands in hyprdots
- menus
- see todos

todo:
- [ ]  Script to handle dynamic stuff
- [ ] Docs
- [ ] Menu to change wayabr layout and style and more++
- [ ] Dynamic borderradisu according to rounding
- [ ] Dynamic shadows according to compositor
- [ ] Port hyprdots and other layouts and style
- [ ] Have fun!

why?
- hyprdots waybar configuration is too limiting.
- Also giving users easy interface like the config.ctl is practically okayish, but learning  json/c a  is much better and opens more window of ideas.  To compensate with that a selector will be used to naviagte through layouts and styles.
- New imple adheres to wayabr documentation and some gtk stuff which you search engine and your A! can answer for you.

# Pull Request

[es](PULL_REQUEST_TEMPLATE.es.md)
[de](PULL_REQUEST_TEMPLATE.de.md)

## Description

Please read these instructions and remove unnecessary text.

- Try to include a summary of the changes and which issue is fixed.
- Also include relevant motivation and context (if applicable).
- List any dependencies that are required for this change. (e.g., packages or other PRs)
- Provide a link if there is an issue related to this pull request. e.g., Fixes # (issue)
- Please add Reviewers, Assignees, Labels, Projects, and Milestones to the PR. (if applicable)

## Type of change

Please put an `x` in the boxes that apply:

- [ ] **Bug fix** (non-breaking change which fixes an issue)
- [ ] **New feature** (non-breaking change which adds functionality)
- [ ] **Breaking change** (fix or feature that would cause existing functionality to not work as expected)
- [ ] **Documentation update** (non-breaking change; modified files are limited to the documentations)
- [ ] **Technical debt** (a code change that does not fix a bug or add a feature but makes something clearer for devs)
- [ ] **Other** (provide details below)

## Checklist

Please put an `x` in the boxes that apply:

- [ ] I have read the [CONTRIBUTING](https://github.com/prasanthrangan/hyprdots/blob/main/CONTRIBUTING.md) document.
- [ ] My code follows the code style of this project.
- [ ] My commit message follows the [commit guidelines](https://github.com/prasanthrangan/hyprdots/blob/main/CONTRIBUTING.md#git-commit-messages).
- [ ] My change requires a change to the documentation.
[...]

PR #345, opened 2025-03-08
Builtin pyutils

kRHYME7:
## Added wrappers and utils for unified python scripts

Features:

###  Added Dynamic pip venv pip installation .
 This is to comply to future implementation of Hydebar/hydepanel. I want it to be installed in a virtual environment to avoid mangling with the system.

 HyDEs requirments.text will be inside ./pyutils/*. Includes `loguro` for  logging and `hyprpy` for hyprland pythonic interface. 

### Added native wrappers:
includes:
1. libnotify to have a pythonic notify-send method
2. fzf for implementing TUI prompts
3. rofi for easier rofi command handling

This is not a breaking implementation, But this serves as backbone implementation for #215 

### Unified Logging

- Uses loguru if installed and fallback to logging
- `LOG_LEVEL=debug` to enable debugging

# Ref

https://github.com/HyDE-Project/HyDE/discussions/236

As requested by @dieBakterie 

# Required by
#215

# Preview

https://github.com/user-attachments/assets/aa02774a-e0ac-4bd2-83b0-feaaf3fdd011

### DIFF
--- a/Configs/.local/lib/hyde/parse.config.py
+++ b/Configs/.local/lib/hyde/parse.config.py
@@ -1,41 +1,13 @@
 #!/usr/bin/env python
 import tomllib
 import argparse
-import logging
 import os
 import time
 import threading
 import subprocess
+import pyutils.logger as logger
 
-
-def fmt_logging():
-    class ColoredFormatter(logging.Formatter):
-        COLORS = {
-            "DEBUG": "\033[94m",  # Blue
-            "INFO": "\033[92m",  # Green
-            "WARNING": "\033[93m",  # Yellow
-            "ERROR": "\033[91m",  # Red
-            "CRITICAL": "\033[95m",  # Magenta
-        }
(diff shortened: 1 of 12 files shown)


dieBakterie:
Does packages like `pyamdgpu` will be added to default requirements too?

kRHYME7:
Ah yeah we can add that.

kRHYME7:
@dieBakterie  By any chance do you have amd gpu to test ?  Maybe we can add a separate PR for this ?

dieBakterie:
> @dieBakterie  By any chance do you have amd gpu to test ?  Maybe we can add a separate PR for this ? 
I don't have an amd GPU

kRHYME7:
@dieBakterie  #346  I realized we can just dynamically install it without adding in requirements as not all is using amd.

dieBakterie (approved):

PR #386, opened 2025-03-21
fix: fix and complete fastfetch.sh implementation

CongLuanTran:
# Pull Request

## Description

Original `fastfetch.sh` implementation is not fully functional. Passing the `logo` commands and its options to restrict the range of random images is not fully implemented. For the old code, the command 
```
fastfetch.sh logo [options]
```
only print the path of a random images according to the options passed to it.

With this fix, `fastfetch.sh` now works as it should be. If called with the `logo` command, it will use the options passed and prints the correct set of images.

## Type of change

Please put an `x` in the boxes that apply:

- [x] **Bug fix** (non-breaking change which fixes an issue)
- [ ] **New feature** (non-breaking change which adds functionality)
- [ ] **Breaking change** (fix or feature that would cause existing functionality to not work as expected)
- [ ] **Documentation update** (non-breaking change; modified files are limited to the documentations)
- [ ] **Technical debt** (a code change that does not fix a bug or add a feature but makes something clearer for devs)
- [ ] **Other** (provide details below)

## Checklist

Please put an `x` in the boxes that apply:

- [ ] I have read the [CONTRIBUTING](https://github.com/prasanthrangan/hyprdots/blob/main/CONTRIBUTING.md) document.
- [ ] My code follows the code style of this project.
- [ ] My commit message follows the [commit guidelines](https://github.com/prasanthrangan/hyprdots/blob/main/CONTRIBUTING.md#git-commit-messages).
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added a changelog entry.
- [ ] I have added necessary comments/documentation to my code.
- [ ] I have added tests to cover my changes.
- [ ] I have tested my code locally and it works as expected.
- [ ] All new and existing tests passed.

## Screenshots

(if appropriate)

PR #387, opened 2025-03-21
fix: complete the implementation of fastfetch.sh

CongLuanTran:
# Pull Request

## Description

Original `fastfetch.sh` implementation is not fully functional. Passing the `logo` commands and its options to restrict the range of random images is not fully implemented. For the old code, the command 
```
fastfetch.sh logo [options]
```
only print the path of a random images according to the options passed to it.

With this fix, `fastfetch.sh` now works as it should be. If called with the `logo` command, it will use the options passed and prints the correct set of images.

## Type of change

## Checklist

- [ ] My code follows the code style of this project.

## Screenshots

(if appropriate)

### DIFF
--- a/Configs/.local/lib/hyde/fastfetch.sh
+++ b/Configs/.local/lib/hyde/fastfetch.sh
@@ -2,13 +2,13 @@
 
 # Early load to maintain fastfetch speed
 if [ -z "${*}" ]; then
-  clear
-  exec fastfetch --logo-type kitty
-  exit
+	clear
+	exec fastfetch --logo-type kitty
+	exit
 fi
 
 USAGE() {
-  cat <<USAGE
+	cat <<USAGE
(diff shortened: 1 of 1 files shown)


kRHYME7:
You want to show fastfetch with the logo right when `fastfetch.sh logo`?

see
https://github.com/HyDE-Project/HyDE/pull/386#issuecomment-2743998261

CongLuanTran:
Closing this commit as well since it is not solving anything.

kRHYME7:
> fastfetch.sh (without argument) a bit redundant

Correct!  But I did follow how hyprlock.sh works also just in case we add more functionality. Letting it run fastfetch without args is faster.And faster is the essence of fastfetch that is why I did not add arg parsing for it. 


Maybe we should modify the description for the logo command so that it won't be misunderstood.

PR #396, opened 2025-03-23
fix: fix the package manager alias

CongLuanTran:
# Pull Request

## Description

Some package manager aliases in `.zshenv` were using the wrong `pm.sh` command and so will not work. I fixed those to the correct `pm.sh` command. I also make the `in` alias to open an interactive panel for choosing package when it is called without any argument. This makes `in` to be more parallel to `un`.

## Type of change

## Checklist

### DIFF
--- a/Configs/.zshenv
+++ b/Configs/.zshenv
@@ -62,34 +62,6 @@ function load_zsh_plugins {
     [[ -r $ZSH/oh-my-zsh.sh ]] && source $ZSH/oh-my-zsh.sh
 }
 
-# Install packages from both Arch and AUR
-function in {
-    local -a inPkg=("$@")
-    local -a arch=()
-    local -a aur=()
-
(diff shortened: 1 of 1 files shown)


kRHYME7:
No need to handle in. 

rubning $PM install will open an fzf menu.

CongLuanTran:
> No need to handle in.
> 
> running $PM install will open an fzf menu.

But in the current set up, the `in` will not be run at all if `inPkg` is empty

kRHYME7:
ah we can deprecate that function.


We can simply have
```
alias in='$PM install`

```

It will be handled by the script.

kRHYME7:
@CongLuanTran  tested? also ready for merging?

CongLuanTran:
> @CongLuanTran tested? also ready for merging?

yeah, it works well and I think it is ready to merge for this file.

kRHYME7:
Thanks!

PR #417, opened 2025-03-28
Parallelize pkg_installed

Jax-Tsai-zero:
**Check three at the same time to save time.**

In resent update, pkg_installed was called in rofi launch script, and it is very slow when uwsm is not installed, so I parallelize it to deal with the lag.

### DIFF
--- a/Configs/.local/lib/hyde/globalcontrol.sh
+++ b/Configs/.local/lib/hyde/globalcontrol.sh
@@ -166,16 +166,12 @@ fi
 #// extra fns
 
 pkg_installed() {
-    local pkgIn=$1
-    if hyde-shell pm.sh pq "${pkgIn}" &>/dev/null; then
-        return 0
-    elif pacman -Qi "flatpak" &>/dev/null && flatpak info "${pkgIn}" &>/dev/null; then
-        return 0
-    elif command -v "${pkgIn}" &>/dev/null; then
-        return 0
-    else
-        return 1
-    fi
+    local pkgIn="$1"
+    parallel --halt now,success=1 --jobs 3 '{}' ::: \
+      "hyde-shell pm.sh pq '${pkgIn}' &>/dev/null" \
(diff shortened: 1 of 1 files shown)


kRHYME7:
Hi! TBH parallel is not a good solution, initializing parallel takes more time than a single `command -v ` option therefore let's preserve the sequential `pkg_installed ` method  and put the command -v on top to make it faster. 

image


# test 

```
#!/usr/bin/env bash

pkg_installed() {
    local pkgIn="$1"
    parallel --halt now,success=1 --jobs 3 '{}' ::: \
        "hyde-shell pm.sh pq '${pkgIn}' &>/dev/null" \
        "( pacman -Qi flatpak >/dev/null 2>&1 && flatpak info '${pkgIn}' &>/dev/null )" \
        "command -v '${pkgIn}' &>/dev/null" &&
        return 0 || return 1
}

pkg_installed_sequential() {
    local pkgIn=$1
    if command -v "${pkgIn}" &>/dev/null; then
        return 0
    elif pacman -Qi "flatpak" &>/dev/null && flatpak info "${pkgIn}" &>/dev/null; then
        return 0
    elif hyde-shell pm.sh pq "${pkgIn}" &>/dev/null; then
        return 0
    else
        return 1
    fi
}

"${@}"

```

verdict: parallel runtime causes latency, though it is consistently 200-300 ms that's huge! So we should just prioritize the ` command -v` in the logic stack.

kRHYME7:
image

This should do it.

Jax-Tsai-zero:
You are right, but the reason why I do this is to accelerate the case when the PKG in not installed,when it not only run command -v but also the two other command after that. So the last change just don't do anything in this case.😂

kRHYME7:
> Or maybe we should change the rofi script, replace the pkg_installed with a single command -v for uwsm?

nope about uwsm it's a special case. We should check if the session is actually handled by uwsm. 


but I think uwsm app -- {cmd} || {cmd}

is enough, if it fails then run command.

Jax-Tsai-zero:
Got it. I'll test it.

PR #420, opened 2025-03-29
Optimize rofilaunch script

Jax-Tsai-zero:
Remove pkg_installed check, and change rofi command, use sh to avoid rofi throw out the error of command not found.

Tested, ready to merge.


kRHYME7:
@Jax-Tsai-zero  Tried on all this flags?

Jax-Tsai-zero:
> @Jax-Tsai-zero  Tried on all this flags? 
> 
> 

I don't see your image, I try to open app using new script and it works fine, and the launch speed is obviously faster.

PR #429, opened 2025-04-01
refactor: style match emoji-picker.sh with glyph-picker.sh

mokeWe:
# Pull Request

[es](PULL_REQUEST_TEMPLATE.es.md)
[de](PULL_REQUEST_TEMPLATE.de.md)
[fr](PULL_REQUEST_TEMPLATE.fr.md)

## Description

Refactor emoji-picker to match the style of glyph-picker due to their similar functionality. Improves code readability and maintainability, and migrates variables to snake_case.

## Type of change

- [x] **Technical debt** (a code change that does not fix a bug or add a feature but makes something clearer for devs)

## Checklist

### DIFF
--- a/Configs/.local/lib/hyde/emoji-picker.sh
+++ b/Configs/.local/lib/hyde/emoji-picker.sh
@@ -77,11 +77,6 @@ parse_arguments() {
             use_rofile=${2}
             shift
             ;;
-        --deps)
-            # nonexistent function?
-            resolve_deps
-            exit 0
-            ;;
         -*)
             cat <<HELP
 Usage:
--- a/Configs/.local/lib/hyde/emoji-picker.sh
+++ b/Configs/.local/lib/hyde/emoji-picker.sh
@@ -1,73 +1,89 @@
 #!/usr/bin/env bash
 
-# Set variables
-scrDir=$(dirname "$(realpath "$0")")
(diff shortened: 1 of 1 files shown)


Configs/.local/lib/hyde/emoji-picker.sh
```diff
@@ -1,73 +1,89 @@
...
-        shift
-        ;;
-    --deps)
-        resolve_deps
-        exit 0
-        ;;
-    -*)
-        cat <<HELP
+            ;;
+        --deps)
```
kRHYME7: yeah the ` --deps` is actually from the Hyde-CLI function which try to check for missing dependencies/commands. You can nuke that for now. 


I am thinking in the future should we add `--deps ` or ` --check`  to some scripts? This will avoid runtime latency as when errors occured the user can call ` --check` and would output diagnostics like installing the command or maybe a notification too if required.

kRHYME7 (changes requested):

Configs/.local/lib/hyde/emoji-picker.sh
httpsworldview: Okay, I'll just remove it for the time being. Something like `--check` would be good :+1:

kRHYME7 (approved):
Tested and LGTM 

image

kRHYME7 (approved):

PR #437, opened 2025-04-03
Use xdg-user-dir to respect user-defined XDG directories

Metalhearf:
# Use xdg-user-dir to respect user-defined XDG directories

## Description

This pull request addresses an issue with HyDE's handling of user-defined XDG directories, i.e. the directory for pictures (`XDG_PICTURES_DIR`). Previously, the configuration used hardcoded paths (like `$HOME/Pictures`), which ignored user customizations that may have been defined in the `~/.config/user-dirs.dirs` file. 

### The Problem with a custom "/home/user/Images" Directory
The issue I faced was that I had customized my `XDG_PICTURES_DIR` to point to `/home/user/Images` instead of the default `Pictures` directory. This was done by editing the `~/.config/user-dirs.dirs` file, which is where users can define custom locations for various directories (Documents, Pictures, Music, etc.).

However, in the previous version of HyDE, the path to the pictures directory was hardcoded as `$HOME/Pictures` in the `Configs/.zshenv` file. Because of this, even though I had set `XDG_PICTURES_DIR` to `/home/user/Images` in my `user-dirs.dirs` file, the system would still reference `$HOME/Pictures`, which was not the directory I wanted to use. This caused a mismatch and the pictures were not being stored in the intended location. (I've found this issue while using the screenshot tool, which kept sending images to `$HOME/Pictures/Screenshots`).

### Motivation for the Change
To address this, I modified the configuration so that HyDE would respect the user's custom directory settings. This solution uses the `xdg-user-dir` command, which automatically fetches the user-defined directory paths, ensuring that the correct directories are used based on what is configured in `~/.config/user-dirs.dirs`.

By making this change, my issue went away. Now, HyDE correctly uses the customized directories like `XDG_PICTURES_DIR`, regardless of whether they are set to the default or custom values, such as my `/home/user/Images`.

## Type of change

## Checklist

## Additional context

This change ensures compatibility with users who have customized their directory structure. It dynamically fetches directories like `XDG_PICTURES_DIR` from the user’s environment configuration, making it flexible for any user setup.


kRHYME7:
is xdg-user-dir native to fresh arch install?

kRHYME7:
I don't mind adding it as default. But do debian have xdg-user-dirs as an official package?

@Senshi111  Will this break debian ?

@richen604  How about in nix? 


This should add another "safe" layer on xdg stuff so I would like to add it also upstream.

Senshi111:
Debian have

richen604:
> @richen604 How about in nix?

yes, although hydenix/nix configs xdg separately ([i dont use`.zshenv` for xdg](https://github.com/richen604/hydenix/blob/4b377bfac41391d6d7516b3253bc3327f06bea20/hydenix/modules/hm/shell.nix#L110))

in general i like this, @Metalhearf you should also include [fish's config](https://github.com/HyDE-Project/HyDE/blob/dc903aaad67730117a78150d6b75cfd921a67a6c/Configs/.config/fish/hyde_config.fish) in case fish users delete `.zshenv`

kRHYME7:
@Metalhearf  Add check if the package is installed. It might return an error message into the variable.

Scripts/pkg_core.lst
kRHYME7, line 1: I got an Idea. To avoid  issues, I will add this package in advance for the upcoming release., 2 days from now.

Then I will merge this one before the next release so we can put up an announcement.  So maybe remove or resolve this change for now. I will push a change later.

kRHYME7 (changes requested):

Scripts/pkg_core.lst
Metalhearf: Oh, alright got it : I see the package has been removed from here and added on :master.
Feel free to keep me up-to-date on the next changes needed when time comes.

kRHYME7: Will merge this end of this month.

kRHYME7:
Ops I have a major concern. I realized we will do a multiple subshell calls. Are we sure to add this into the init of phase of the shell?

kRHYME7:
@Metalhearf Will you clean up the PR and point this to the dev Branch. zsh now is using user-dirs, So it we only need fish support

Metalhearf:
Alright @kRHYME7, I’ve added the missing parts and removed the `.zshenv` changes as requested.

I’m not a `fish` user, so I tested this as best I could with basic source/env/echo shenanigans. If someone using fish could verify it looks alright on their side, that’d be nice. :)

Thanks!

PR #445, opened 2025-04-06
[enhancement]: Speed up install_pkg.sh

Jax-Tsai-zero:
Speed up install_pkg.sh from 5.2s to 390 ms.

image


kRHYME7:
Do we really needed this speed than being more accurate?  It's an installation script. 

Been using bash for quite some time now, and mapfile + this (IMG) could cause unexpected issues. 
image
I don't have a solid proof except for the bugs we got here in HyDE that circles around mapfile + ` ~=` operators.

kRHYME7:
We can use ` -Qq ` flag though to at least shred some speed. X2 better 
image

Jax-Tsai-zero:
> We can use `-Qq` flag though to at least shred some speed. X2 better image

It's good. but actually I don't get the "unexpected issues"🤔. Could you just explain it? It works fine on my device.

I think maybe the speed of the install script can be optimized cuz every time we install or update, it takes more then 40 sec. (on my device).
btw I think the tui of the install script can be improved, maybe gum and tput is good. What's your opinion?

kRHYME7:
>  btw I think the tui of the install script can be improved, maybe gum and tput is good. What's your opinion?


MR would be appreciated. 


> I don't get the "unexpected issues"🤔. Could you just explain it? 

- I noticed that you're using unquoted string, which means if a package name in the `install_pkg.lst` is problematic or more generalized it will break the logic
image

Though we can use safer methods by matching array[@] with "string", this means we are adding our own abstractions which is a burden to maintain. 

> It works fine on my device.

bad mindset haha

Jax-Tsai-zero:
@kRHYME7 You're right. Thanks for the patient expaination. 🙏

kRHYME7:
It's  good to learn something together!

PR #477, opened 2025-04-12
feat: Screen Recorder script

0xAquora:
# Pull Request

## Description

Motivation:
Sometimes i want to record a quick video explaining how to do something and opening OBS is too much hassle for such a small task so i decided to implement a few keybindings to make recording videos more comfortable for the users.

Description:
Its a little script that checks the encondings allowed by your gpu and uses slurp to choose the region you want to share, afterwards it gets saved in /Videos.

Dependencies:
[wf-recorder](https://github.com/ammen99/wf-recorder)

## Type of change

## Checklist


kRHYME7:
I don't know, maybe learn how we structure our scripts first?

If it's specific to the wfrecorder, then maybe just put it into a single wfrecorder.sh script? 



Fun fact `Hyde-CLI` had a script for this would you like me to port this to HyDE? It supports 2 backends wfrecorder and wlscreenrec.

kRHYME7:
Would you like me to add it for you?

0xAquora:
> Would you like me to add it for you?

Yes please, that would be great!

Also it is true that it all could have fit into a single file commit, i am sorry.

kRHYME7:
> Also it is true that it all could have fit into a single file commit, i am sorry. 


Yup

0xAquora:
I merged the functionality of those 3 files into wf-recorder.sh as adviced and updated keybindings accordingly

kRHYME7:
Do you have any reasons not using wl-screenrec?

0xAquora:
Because i am unsure if it works properly with Nvidia GPUs following this image, so since theres different tools for GPUs depending on which specific GPU you've got, i preferred to go with what i saw to be more standarized.

Though if you have any suggestion, i can try to make a script which chooses preferred tool for recording, but i think that should be declared on the installation script to not bloat the user.
image

TowhidMolla-Dev:
update?

Configs/.config/hypr/keybindings.conf
kRHYME7, line 1: Can we remove the binds for now @0xAquora?

Scripts/pkg_core.lst
kRHYME7, line 1: move this to pkg_extra.lst  too.

kRHYME7 (changes requested):

Configs/.config/hypr/keybindings.conf
0xAquora: Yeah ofc

0xAquora:
Changes done, mister 🫡

kRHYME7:
It fails. Do we really need to explicitly add the encoders? I this wf-recorder handles it automatically

image

kRHYME7:
# USAGE

` screenrecord.sh` 
BTW ` hyde-shell` calls the scripts in ` hyde lib` which is convenient when scripting or doing some other interopt. 

image

## Basics for binds


To start
``` 
hyde-shell screenrecord --start 
```

To exit/quit 
```
hyde-shell screenrecord --start 
```

To use either wl-screenrec or wf-recorder, we can add `--backend <backend>`

kRHYME7:
@TowhidMolla-Dev  @0xAquora  Please add your review. this should work for both wl-screenrec and wf-recorder.

Configs/.local/lib/hyde/screenrecord.sh
```diff
@@ -0,0 +1,151 @@
...
+        split($1, pos, ",");  # Split x,y
+        x = pos[1];
+        y = pos[2];
+        split($2, size, "x");  # Split WxH
+        width = size[1];
+        height = size[2];
+        
+        # Check if selection meets minimum size
+        if (width >= 16 && height >= 16) {
+            print x","y" "width"x"height;  # Output in the same format
```
dieBakterie, line 91: Are you sure this line is correct, seems like an `"` is missing?

0xAquora (commented):
Works as expected

Configs/.local/lib/hyde/screenrecord.sh
kRHYME7: Yes that's awk pipeline to handle crazy math in bash without adding dependency.

in awk, the quotes are used to concatenate string and x , y width and height are variables.
quoted comma "," , quoted space " " ,and quoted x "x" for the multiplication symbol, were used to output a pseudo  slurp command output 

```
1234,1234 123x123

```

kRHYME7:
Merging this now, thank you @0xAquora

PR #479, opened 2025-04-14
Add workflow mode with 5 persistent modes.

JunaidQrysh:
# Pull Request

## Description

The problem is some users dont have beefy PCs or they just want to extend their laptop battery or have different preferences like - no borders, no opacity, (no border with opacity & blur), etc.

This MR solves that by introducing 5 different combinations of workflow modes(or preferences) 

https://github.com/user-attachments/assets/958773f3-2804-4d16-a0eb-4c8df4c202de

## Type of change

## Checklist

- [ ] I have read the [CONTRIBUTING](https://github.com/prasanthrangan/hyprdots/blob/main/CONTRIBUTING.md) document.
- [ ] My commit message follows the [commit guidelines](https://github.com/prasanthrangan/hyprdots/blob/main/CONTRIBUTING.md#git-commit-messages).

## Screenshots
250414_14h54m31s_screenshot
Each icon representing different mode
250414_14h57m30s_screenshot

### DIFF
--- a/Configs/.local/lib/hyde/workmode.sh
+++ b/Configs/.local/lib/hyde/workmode.sh
@@ -1,0 +1,82 @@
+#!/usr/bin/env sh
+
+MODE_PATH="$HOME/.cache/workflow"
+TMP_PATH="$XDG_RUNTIME_DIR/workflow"
+
+if [ "$1" = "g" ]; then
+    mode=$(cat "$MODE_PATH")
+    case $mode in
+        0) echo "󰸶" ;;
+        1) echo "󰸸" ;;
+        2) echo "󰸷" ;;
+        3) echo "󰸴" ;;
+        4) echo "󰸵" ;;
+    esac
+    exit 0
+fi
+
+cycle_mode() {
+    local mode=$1
+    mode=$((mode + 1))
+    [ $mode -gt 4 ] && mode=0
+    echo $mode
+}
+
+update_mode_files() {
+    local mode=$1
+    echo $mode >"$TMP_PATH"
+    echo $mode >"$MODE_PATH"
+}
+
+if [ ! -f "$TMP_PATH" ]; then
+    mkdir -p "$(dirname "$MODE_PATH")"
+    [ ! -f "$MODE_PATH" ] && echo 0 >"$MODE_PATH"
+    mode=$(cat "$MODE_PATH")
+else
+    mode=$(cat "$TMP_PATH")
+    [ $# -eq 0 ] && mode=$(cycle_mode "$mode") && hyprctl reload
+fi
+
+update_mode_files "$mode"
+
+case $mode in
+0)  
+    ;;
+1)
+    hyprctl --batch "keyword decoration:blur:enabled 0; \
+                     keyword decoration:shadow:enabled 0; \
+                     keyword decoration:active_opacity 2; \
+                     keyword decoration:inactive_opacity 2"
+    hyprctl 'keyword windowrulev2 opaque,class:(.*)' # ensure all windows are opaque
+    ;;
+2)
+    hyprctl --batch "keyword animations:enabled 0; \
+                     keyword decoration:blur:enabled 0; \
+                     keyword decoration:shadow:enabled 0; \
+                     keyword general:gaps_in 0; \
+                     keyword general:gaps_out 0; \
+                     keyword general:border_size 0; \
+                     keyword decoration:active_opacity 2; \
+                     keyword decoration:inactive_opacity 2; \
+                     keyword decoration:rounding 0"
+    hyprctl 'keyword windowrulev2 opaque,class:(.*)' # ensure all windows are opaque
+    ;;
+3)
+    hyprctl --batch "keyword animations:enabled 0; \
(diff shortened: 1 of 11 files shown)


kRHYME7:
You can just add the workmode and leave gamemode alone. Let's keep it as that Script is from the official wiki and people are kinda used to it.

kRHYME7:
Also fix your PR description.

Configs/.local/lib/hyde/workmode.sh
```diff
@@ -0,0 +1,83 @@
+#!/usr/bin/env sh
+
+MODE_PATH="$HOME/.cache/workflow"
+TMP_PATH="/tmp/workflow"
```
kRHYME7: Use xdg_runtime_dir for this

Configs/.local/share/hyde/hyprland.conf
```diff
@@ -15,6 +15,8 @@
 
 $scrPath=$HOME/.local/lib/hyde # set scripts path
 
+exec-once = $scrPath/workmode.sh f # set work mode
+
```
kRHYME7, line 19: No, use waybar to launch it for now. 

Assuming that the only interface that is available is the waybar. This way if the waybar reloads this reloads also with corresponding visuals. 

Loading it in a background without something to view is hard to debug/remember.

JunaidQrysh:
but they are the same thing, you just have to press two times and you will have gamemode
or should i make at the first press

kRHYME7 (changes requested):

kRHYME7:
> but they are the same thing, you just have to press two times and you will have gamemode
or should i make at the first press

Doesn't matter. We can't just deprecate old stuff because a cooler one emerges. 

Deprecation should be discussed in a separate thread. Please bring it back for now.

Configs/.local/share/hyde/hyprland.conf
JunaidQrysh: but workmode should be independent of waybar because it is persistent across reboots
and i have been using the script for weeks now and not a single issue.

kRHYME7: Isn't waybar doing the same thing? Though it is an agnostic script, our only interface to query is the persistent file and waybar's icon.

Now for example we got issues with hyprland and these hyprctl commands are being loaded (without waybar), do you expect us to always remember that this script is running?

hyprctl commands are loaded in memory and we got a bunch of configuration to be able to pin point what to query  that is running in the background. 

Another note, Hyprland's exec* calls are bottlenecks, letting this slide in means I'll be accepting more. If waybar is capable of launching this then why not?

JunaidQrysh: ok, i will do that, but on what layouts should i include the module
and is Super+Shift+G good for manually launching the script?

kRHYME7: > ok, i will do that, but on what layouts should i include the module

You can change all gamemode into workmode, 


> and is Super+Shift+G good for manually launching the script?

Isn't there a bind for gamemode.sh? use that for workmode.


My only concern about preserving gamemode is the module and the script should stay for now. We don't know if someone is using it in their keybindings.conf or in their waybar layouts.

Configs/.local/lib/hyde/workmode.sh
```diff
@@ -0,0 +1,82 @@
...
+4)
+    hyprctl --batch "keyword animations:enabled 0; \
+                     keyword general:gaps_in 0; \
+                     keyword general:gaps_out 0; \
+                     keyword general:border_size 0; \
+                     keyword decoration:rounding 0"
+    ;;
+esac
+
+pkill -RTMIN+10 waybar
```
kRHYME7, line 82: Hmmm I realized something , what happened to the configs that are not being unset or overwritten?

Configs/.local/share/hyde/hyprland.conf
JunaidQrysh: ok, what about persistence if manually triggering the scirpt?

kRHYME7: Will check the whole functionality when I get back to  my laptop.

 Please correct the PRs description first, it's hard to read the script on my phone.

kRHYME7 (changes requested):

kRHYME7:
Now I realize we can make it even more customizable! 

what i Invision is something like a set of configurations where they can choose.

example. 
```
workmode.conf 
gamemode.conf
editors.conf
```

then we can just do an IPC or use hyprctl.

`hyprctl -r -- keyword source $HYPR_MODE`


I'll be experimenting on this approach as I can see this is suited for hyrpland very well.

kRHYME7:
If you can check the /animations/ we can replicate that and create a set of configurations for each workflow.

kRHYME7:
@JunaidQrysh  Check this out https://github.com/HyDE-Project/HyDE/pull/480

JunaidQrysh:
Great!
Now you get the idea

kRHYME7:
>  So, should i close this MR?

Yup and let's continue at #480

Issue #501, opened 2025-04-17
[Bug]: Font Resolution Fails Due to Tilde Expansion in `font.sh`


0xGeN02:
## Description:

When invoking the font.sh resolve command with a layout path specified as `~/.config/hypr/hyprlock/theme.conf`, the script fails to locate the file. This is because tilde expansion ```~``` is not performed within double quotes, leading to the literal interpretation of ```~``` instead of expanding it to the user's home directory ```$HOME```.

![Image](https://github.com/user-attachments/assets/e58f21e2-46dc-4cfb-8cfd-58a0024cba55)

## Steps to Reproduce:

1. Set the `layout path` in `~/.config/hypr/hyprlock.conf` as:

```sh
$LAYOUT_PATH=~/.config/hypr/hyprlock/theme.conf
```

2. Run `font.sh resolve` with the layout path:

```sh
font.sh resolve "$LAYOUT_PATH"
```

## Expected Behavior:

The script should correctly expand ~ to the user's home directory and locate the theme.conf file.​

## Actual Behavior:

The script fails to find the file because ~ is not expanded inside double quotes, resulting in an incorrect path.

## Solution:

Replace the tilde (~) with $HOME to ensure proper expansion within double quotes:

```sh
font.sh resolve "$HOME/.config/hypr/hyprlock/theme.conf"
```

### File pre changes:

![Image](https://github.com/user-attachments/assets/9266da11-2739-4ac9-8fec-d8d70a5f08c7)

### File post changes:

![Image](https://github.com/user-attachments/assets/5a161c04-c16f-4e45-a600-ed7295fb7f88)

### Solution output (no error):

![Image](https://github.com/user-attachments/assets/4a7512fe-ef1d-410d-8398-3ec4e503443d)

kRHYME7:
![Image](https://github.com/user-attachments/assets/204e6727-8499-426b-8652-4c8994496301)

A tilde with dqoutes should act as a tilde character and should't be expanded as HOME. Expanding it while being in dqoutes is not a good convention to follow. Better to use $HOME .

Issue #478, opened 2025-04-12
[Bug]: Hyprlock crash while media is playing


SadScientist:
### Description

After just having installed Hyde Hyprlock crashes immediately upon locking the screen **if audio or video is actively playing**. 

The crash appears to be triggered specifically when hyprlock attempts to load or render an `image` widget configured via ~/.local/share/hyde/hyprlock.conf. In this case, commenting "$MPRIS_IMAGE = $XDG_CACHE_HOME/hyde/landing/mpris.png" fixes the issue. The crash doesn't occur in some cases, though haven't been able to determine exactly when. I've tested this with Youtube and Spotify so far.

### Steps to Reproduce

1.  Use default Hyprlock configuration after installing HyDE
2.  Start playing audio using any application (e.g., music player, web browser video).
3.  While the audio is actively playing, trigger `hyprlock` (e.g., via `hyprlock` command in terminal, keyboard shortcut, or idle timer).

### Expected Behavior

Hyprlock should successfully lock the screen and display the song or video name and a thumbnail image.

### Actual Behavior

The lock screen is shown with the thumbnail, but crashes almost immediately.

### Screenshots

It's the Hyprlenol :( screen

### Environment

- Linux Distro: **Arch Linux**
- Linux Version: Linux archlinux 6.12.22-1-lts #1 SMP PREEMPT_DYNAMIC Tue, 08 Apr 2025 06:20:12 +0000 x86_64 GNU/Linux

kRHYME7:
@SadScientist  Does the file exist?

`$XDG_CACHE_HOME/hyde/landing/mpris.png`

SadScientist:
> [@SadScientist](https://github.com/SadScientist) Does the file exist?
> 
> `$XDG_CACHE_HOME/hyde/landing/mpris.png`

It does exist, yes. It's the image I see when no music is playing.

CongLuanTran:
I can confirm that it happens to me too.

CreepySunny:
# Description
I've also encountered this issue. I wasn't able to record a hyprlock error log, the only error logs I got is the following:
```
[  324.456355] sof-audio-pci-intel-tgl 0000:00:1f.3: ipc tx error for 0x60010000 (msg/reply size: 108/20): -22
[  324.456377] sof-audio-pci-intel-tgl 0000:00:1f.3: HW params ipc failed for stream 1
[...]

PR #502, opened 2025-04-17
refactor: font.sh from (~) to ($HOME)

0xGeN02:
# Solve hyprlock font importation

see #501


0xGeN02:
this may help also with #478 in last 5 ```hyprlock``` commands done via shell while fixing the issue I had 0 crashes

kRHYME7 (changes requested):
This is bad.

Just use` realpath ` for every directory in this script. Also can you provide a scenario that this script fails?

kRHYME7:
latter query is a separate issue

0xGeN02:
> I fixed the path error.
> 
> Also the contents of this? image
> 
> Be sure that `curl -s -O -L "${url}"` can handle your URL. But by the error your font2.sh has syntax errors in it.

yes there was a typo on my .layout file where I try to fetch the .zip

PR #521, opened 2025-04-24
Implement UWSM as the session manager

slapjd:
# Pull Request

## Description

I've seen in issues/discussions that you are open to using UWSM as the session manager for HyDE. I have this set up on my machine so I figured I'd test the waters and submit a PR.

Overarching summary is as follows:
1. Environment management is entirely moved to UWSM. `~/.config/uwsm/env-hyprland` sources `~/.config/uwsm/hyde/*` which is assumed to not be overwritten by the user. The idea is that `env-hyprland` and `env` can still be used for any custom environment variables (provided they're placed after the hyde sourcing so they're overwritten). Existing variables have been placed in `~/.config/uwsm/hyde/env` (except for nvidia variables, placed in `~/.config/uwsm/hyde/env-nvidia` and only synced if `nvidia-utils` is installed).
2. Because `$scrPath` was put into `PATH` I made a new environment variable `HYDE_LIB_DIR` similar to the other `HYDE_*_DIR` variables to replace it (so it can be put in UWSM environment and follow environment naming conventions).
3. `$start.DBUS_SHARE_PICKER` and `$start.SYSTEMD_SHARE_PICKER` have been obsoleted seeing as UWSM already takes care of it.
4. `uwsm app` is used to run any "normal" program as of right now. It might make more sense to use `app2unit` or `runapp` to reduce startup time. Small short-lived things like the process killer, volume up and down keybind scripts, etc. are just run as they are. Background services are largely left untouched as well, although it might be worth considering running them in the background slice? I wasn't sure whether something like waybar for example would count as a core part of the DE or a background app. The difference large doesn't matter anyway.
5. `uwsm stop` is used to logout instead of `loginctl` because otherwise SDDM will not start after logging out for some reason.

On my machine it runs equivalently to normal HyDE, as far as I can tell. SDDM login/logout works with the uwsm-managed Hyprland entry. There may be some documentation needed to tell people that the uwsm-managed desktop entry needs to be used now but I wasn't sure where to put it.

## Type of change

- [x] **Breaking change** (fix or feature that would cause existing functionality to not work as expected)

## Checklist


kRHYME7:
Overall this is OK. But let me make some gradual changes first. I want to make HyDE work even without UWSM.

---

example the logout part 


Is it best to use `uwsm stop` instead of using loginctl? (Though loginctl terminate-session sometimes kill the sddm-helper).

slapjd:
I think `uwsm stop` is fine for UWSM sessions until the issue with `loginctl` is fixed for SDDM. It should only have a functional difference if you launch hyprland without `exec`ing it, where `uwsm stop` will only stop Hyprland and would drop you back to a shell and `loginctl` would just log you out completely. For non-UWSM the only proper way to shutdown is the Hyprland exit dispatch (as it currently does) afaik because `loginctl` can cause Hyprland to close before all its subprocesses close, causing UI toolkits and other programs to hang.

I didn't like the idea of HyDE setting environment variables when you're not using it (e.g. logging into another DE or something) but I suppose other configs get set globally already so that ship has kinda already sailed (provided it doesn't *overwrite* variables that are already set, especially for XDG). It would be nice if whatever solution for environment outside UWSM did not touch the environment outside HyDE imo but that would be hard without a custom launcher of some sort. Also maybe worth noting is that GDM does not seem to source the login shell profile, nor does it source /etc/profile or anything like that as far as I can tell on my machine. It's not SDDM, so idk if you care but I thought I should mention it if you want to put environment variables into shell profiles.

kRHYME7:
> although I can't see much reason why they would care.


True! But nevertheless,  we should still try to launch script/command assuming that uwsm is not present. (as I've said on earlier comment).


We might have multiple merge conflicts during that phase so I hope you can take care of that. 


Also by that "assuming uwsm is not present", can we add env vars for at least with `zsh` and `fish`.

example `export XDG_CONFIG_HOME=${XDG_CONFIG_HOME:-$HOME/. config} `

kRHYME7:
> Also by that "assuming uwsm is not present", I am planning to add env vars for at least with `zsh` and `fish`. 

Oh about this which comes first? UWSM launching the proper env vars or user's shell environment?

slapjd:
Yeah that makes sense if you're concerned about people wanting to use raw Hyprland. It might be worth considering a desktop entry that runs a script to source the environment (from the UWSM config, which is just shell scripts) and then launch Hyprland instead of just placing it in shell profiles. That way it won't matter what shell users are logging in with as long as a POSIX shell or bash is installed. I can help with the desktop entry if needed (I already did something similar to work around GDM)
As far as I know the login shell should be sourced before UWSM launches. UWSM used to source /etc/profile but no longer does if it sees there is an environment already I believe.
I'm using HyDE anyway with (roughly) this config so I'll rebase and push whenever I update my system. If you need me to specifically do that at any point just gimme a shout

kRHYME7:
Considering this


` shell > uwsm > processes ` workflow. Removing uwsm we will have `shell > processes`. 

1. We cannot add env vars in shell profiles right? as it will conflict uwsm? or can we make the uwsm environment  overwrite shell profiles? So if this is possible, export the env vars in shell like running `export_environment.sh`on shell. Then if uwsm exists or being run then ofc we want it to overwrite the shell profile envs. Having to run  `export_environment.sh`should work across other shells(hope so).


2. Best bet is to add it in the process which is tricky. As `env` in Hyprland cannot really process shell env correctly. Unless we summon vaxx, and ask to make the `env` keyword act like how shell works. e.g. `env = FOO,${BAR:-fallback}` or something convenient.  That's the reason we remove it in this PR right?.
    Though I am curious if ` exec-once = export_environment.sh ` can work?
this script is like a shell profile too. Back in the day `exec-once = export FOO=bar` works! But I don't know if it is possible to this date.


3. Adding desktop entry is one of the solutions I can think of. But I am trying to avoid  system stuff as much as possible. By that, I think this is reasonable as long as other methods in user space won't work.


---

Also, I think asking the uwsm author for his opinion would also help us choose the right track. Maybe we can open up discussions in his repo and ref this?

slapjd:
UWSM can see any variables that it inherits from when it's launched as far as I'm aware. If you were to set a variable in a shell profile, I think the env scripts in UWSM can see it.
What I don't know is if those variables would then get exported to systemd properly. If they are then it is presumably fine to put environment into shell scripts, but if they aren't then apps run through UWSM might not be able to see them, meaning QT themes, etc might be wrong.

kRHYME7:
@slapjd  Should we open a help issue in UWSM? Can you delegate it for me?

slapjd:
I'm at my machine now

>     1. We cannot add env vars in shell profiles right? as it will conflict uwsm? or can we make the uwsm environment  overwrite shell profiles? So if this is possible, export the env vars in shell like running `export_environment.sh`on shell. Then if uwsm exists or being run then ofc we want it to overwrite the shell profile envs. Having to run  `export_environment.sh`should work across other shells(hope so).

Looking at UWSM docs and running a quick test on my machine, environment variables can be exported perfectly fine from the login shell as far as I can see. If you were to set the same variable on login and on UWSM without checking if it's already set, UWSM would overwrite the login one (because it is executed *by* the login shell). It doesn't really matter for us anyway because we'd be setting the same variables to the same values.

Unfortunately, a portable `export_environment.sh` can only work for POSIX shells (bash, zsh) and so will not work on fish. You'd need a special `export_environment.fish` for fish to work because that `export_environment` script needs to be sourced, not called and so it can't switch shell languages (and fish doesn't understand normal shell scripts).

If you are going to set the variables within the shell config, there is no point in setting them again in UWSM anyway, because it'll just receive them from the login shell I think. I still think it would be best to avoid shell configs if possible though because we don't want to set variables like the QT theme or the HyDE specific variables on a session that might not actually be running HyDE (they could be logging in to another DE or even a VT only session and it would still run the login shell config)

>     2. Best bet is to add it in the process which is tricky. As env in Hyprland cannot really process shell env correctly. Unless we summon vaxx, and ask to make the env keyword act like how shell works. e.g. env = FOO,${BAR:-fallback} or something convenient. That's the reason we remove it in this PR right?.
Though I am curious if exec-once = export_environment.sh can work?
this script is like a shell profile too. Back in the day exec-once = export FOO=bar works! But I don't know if it is possible to this date. Or even safe to do lol.

Being able to export an environment variable if it's unset within Hyprland would probably solve the problem, yes. I'm pretty sure it *should be* impossible to set an environment variable by `exec`ing because `exec` creates a new process and the current process' environment can't be changed from inside the new process. Unless Hyprland is doing something weird that i'm not aware of, it *should* be impossible without a special version of `env =` that doesn't overwrite.

>     3. Adding desktop entry is one of the solutions I can think of. But I am trying to avoid system stuff as much as possible. By that, I think this is reasonable as long as other methods in user space won't work.

Adding a desktop entry would be equivalent to making our own session manager for HyDE (one that doesn't involve systemd slices). It would likely just be a small POSIX sh script equivalent to `source ~/.config/uwsm/env-hyprland; source ~/.config/uwsm/env; exec Hyprland` and a corresponding desktop entry. It would prevent us from polluting the global shell environment with variables specific to HyDE even when we aren't running HyDE and would also let us reuse the uwsm scripts so we don't have to write the same variable values in like 3 different places.

I agree that modifying system stuff without having a package isn't great, but as long as we make sure to put the desktop entry (and bash script) into `/usr/local` and not `/usr` it's not a terrible idea imo. The only reason I would want to avoid this is because it is basically just rewriting what UWSM does in the first place to some extent but if systemd-less operation is important then it's not really wrong to do it.

>     Should we open a help issue in UWSM? Can you delegate it for me?

I don't know what you mean by delegating it but I can open a discussion on there and ask what environment variables get exported to systemd if that's what you mean

kRHYME7:
> I don't know what you mean by delegating it but I can open a discussion on there and ask what environment variables get exported to systemd if that's what you mean

Yes. Though I think I understand now.


> Unfortunately, a portable export_environment.sh can only work for POSIX shells (bash, zsh) and so will not work on fish. You'd need a special export_environment.fish for fish to work because that export_environment script needs to be sourced, not called and so it can't switch shell languages (and fish doesn't understand normal shell scripts).

I'm not on my machine but , can't we just `sh export_environment.sh` this? But now I think this is messy! ignore this dumbness.

--- 
 
It's a bad idea to set HyDE specific envs in shell, but can we atleast export XDG specs on shell that HyDE supports. Namely zsh and fish. About the QT envs and some env that only Hyprland/HyDE will benefit. If it's static (we don't usually change), Ig it is safe to put in `env = ` hyprland config?

Though iirc, uwsm supports exporting env vars per session using the <session>.desktop names. 

---

In conclusion and what I know:

- xdg dirs  can be added in the shell and be resolved with tools like `xdg-user-dirs`. And I think .local/lib ,.cache aren't translatable for xdg so safe to hard code.

-  uwsm handles xdg automatically too which I think no need for a config?


Now, what we needed to consider  are HyDE specific envs.

- Uwsm can  use per session env vars. (I forgot the Syntax but I'm sure it exists). This way we can filter out HyDE specific env vars. Assuming users have uwsm.

- Now how about users without uwsm, I hope that `env ` handles the env correctly. Shall we investigate Hyprland for this?

 Example  in Hyprland config.
`env = VAR,${VAR:-fallback}`. 

❌Either Hyprland treats this as a string leading to `VAR=${VAR:-fallback}`. 

✅ or if no Uwsm or var is not yet set
`VAR=fallback

✅ with uwsm:
`VAR=valueFromUWSM` 


❌ - Not Ideal
✅ - Ideal

If Hyprland  cannot handle the shell syntax yet. Our only option is to use UWSM for HYDE specific env. Then for non uwsm users we can provide a `environment.conf` file with all the necessary Env vars. How about that? This can be done manually or automatically.

slapjd:
> I'm not on my machine but , can't we just `sh export_environment.sh` this? But now I think this is messy! ignore this dumbness.

If you run `sh export_environment.sh` from any shell your environment won't change. It will spawn a new `sh` process, change the environment of *that*, then promptly discard the `sh` process and its environment because it finished running. A child process can't modify a parent process' environment variables, as far as I know. `source export_environment.sh` on the other hand would tell your *current* shell to read and interpret everything in `export_environment.sh` as if you had just typed it in, meaning any `export VAR=blahblah` would be executed in your current shell but fish would interpret it as fish syntax, which might not be valid.

> It's a bad idea to set HyDE specific envs in shell, but can we atleast export XDG specs on shell that HyDE supports. Namely zsh and fish. About the QT envs and some env that only Hyprland/HyDE will benefit. If it's static (we don't usually change), Ig it is safe to put in `env = ` hyprland config? and should avoid adding it in shell profile.

using `env = ` will only work for systems without UWSM, otherwise systemd won't get the updated variables properly (and so apps run with `uwsm app` won't get them I think).

> Though iirc, uwsm supports exporting env vars per session using the .desktop names.

It does, which is why I source the HyDE environment from `env-hyprland`.

> In conclusion and what I know:
> 
>     * xdg dirs  can be added in the shell and be resolved with tools like `xdg-user-dirs`. And I think .local/lib  aren't translatable for xdg so safe to hard code.
> 
>     * uwsm handles xdg automatically too which I think no need for a config?

If you create XDG variables during shell login, UWSM won't overwrite them I would assume. If they are unset UWSM will take care of them. It's not harmful to set them before UWSM starts though, so yes they can be put in the shell config. I *think* xdg-user-dirs only relates to directories like Documents, Music, Downloads, etc? The variables we need are for the XDG Base directories, which I'm pretty sure are exclusively set through environment variables (or the values are assumed to be well-known defaults, like $HOME/.config).

> Now, what we needed to consider are HyDE specific envs.
> 
>     * Uwsm can  use per session env vars. (I forgot the Syntax but I'm sure it exists). This way we can filter out HyDE specific env vars. Assuming users have uwsm.
> 
>     * Now how about users without uwsm, I hope that `env ` handles the env correctly. Shall we investigate Hyprland for this?
> 
> 
> Example in Hyprland config. `env = VAR,${VAR:-fallback}`.
> 
> ❌Either Hyprland treats this as a string leading to `VAR=${VAR:-fallback}`.
> 
> ✅ or if no Uwsm or var is not yet set `VAR=fallback
> 
> ✅ with uwsm: `VAR=valueFromUWSM`
> 
> ❌ - Not Ideal ✅ - Ideal
> 
> If Hyprland cannot handle the shell syntax yet. Our only option is to use UWSM for HYDE specific env. Then for non uwsm users we can provide a `environment.conf` file with all the necessary Env vars. How about that?

I'm like 80% sure Hyprland can't do things like `VAR=${VAR:-fallback}` but I'll check. An `environment.conf` could potentially work if we had some way of detecting that we need it? Maybe if we set a variable like `export HYDE_UWSM=1` in UWSM specific env and then run an `exec-once = /path/to/some/script.sh` in Hyprland, which will run something like `if HYDE_UWSM == 1; symlink /path/to/environment.conf $XDG_CONFIG_HOME/hypr/environment.conf; fi`. We can then have `source = $XDG_CONFIG_HOME/hypr/environment.conf` in Hyprland in the funny block that ignores errors so it sources it if it's there.

It's a little bit convoluted but it *does* solve the problem cleanly (although it might negate any startup time benefit from dropping systemd in the first place lol). Or, as you said, we can just have it be a manual config somewhere if necessary.

kRHYME7:
> It's a little bit convoluted but it does solve the problem cleanly (although it might negate any startup time benefit from dropping systemd in the first place lol). Or, as you said, we can just have it be a manual config somewhere if necessary.


I tried it and doesn't work. Manual for now.

Gabz4200:
Leaving my 2 cents as someone very interested in using UWSM as the session manager with HyDE:

It might be a good idea to consider replacing the current `uwsm app --` commands with `app2unit --`.

According to the [UWSM README](https://github.com/Vladimir-csp/uwsm), **app2unit is a faster, feature-parity shell alternative to `uwsm app`**, created by the same author. It also works independently of UWSM and can integrate with launchers.

Additionally, since you mentioned **"I want to make HyDE work even without UWSM"**, this change would help. app2unit doesn't require UWSM, it only depends on `systemd` and `xdg-terminal-exec`, both of which HyDE already uses.

One possible setup could be:

As seen in [app2unit README](https://github.com/Vladimir-csp/app2unit), we can set this var

```bash
export APP2UNIT_SLICES='a=app-graphical.slice b=background-graphical.slice s=session-graphical.slice'
```

So we could put it in `~/.config/uwsm/env`, and then switch all app launches on HyDE to use `app2unit`.

The thing here is that **everything becomes a managed systemd unit**, but without adding significant overhead, since **app2unit is fast**. So probably no one would complain with or without UWSM.

Just wanted to share this in case it helps shape the final implementation. Thanks for the awesome work on this project!

Btw, English is not my first language, so sorry if I made any mistakes here

kRHYME7:
Thanks @Gabz4200  Will try to look into this

kRHYME7:
BTW, does xdg-terminal-exec protocol finally got merged?

Gabz4200:
I looked into it, and it seems the xdg-terminal-exec, as in the protocol proposal, is still not officially merged into upstream (or at least I couldn't find it).

On the other hand, the CLI tool implementation by Vladimir (https://github.com/Vladimir-csp/xdg-terminal-exec) works well and I actually use it.

app2unit has this too "app2unit -T or app2unit-term (without command) can be used to open default terminal as a unit, with unit metadata filled from its desktop entry.", it does the xdg-terminal-exec thing plus wraps it in an unit

So, protocol-wise, not yet
CLI/tool-wise, already usable.

deepu105:
Hey folks, is this branch usable or is it still WIP?

Gabz4200:
Hi @deepu105 , as far as I can tell, it is usable as long as you stick to UWSM, but it doesn't make the branch less of a WIP. 

There is just one tiny problem, it is very outdated. So you, if you merged it on your local HyDE git file (supposing you have one), you would have to, first, solve the merge conflicts that will happen on some files, then be sure to move any env declaration that still on the hyprland configuration to the uwsm env file that appeared on Configs/.config/uwsm/, since the management is different. 

If you are worried about updates on the HyDE side regards envs, you can set "exec-once = uwsm finalize [ALL_ENV_VAR_NAMES_YOU_WANT]" at the end of .config/hypr/hyprland.conf, or, if you don't set any envs on userprefs, go to hyde.conf and change "SYSTEMD_SHARE_PICKER" to this same uwsm command, "uwsm finalize" does the same as the dbus-activation and systemd user import-envs, plus some uwsm things.

It may be boring if you don't know where to look at, but it works.

I think there are still some scripts on .local/lib/hyde that are not using uwsm, like hyde-launch that uses gtk-launch, but these don't break anything, it just doesn't use the full potential of uwsm.

Maybe I can make my own pull request with these fixes... But I am not confident enough to do it and don't feel I made something that is not good enough :p

Gabz4200:
I am writing in a rush (like in being very fast because I am late, I don't know if this is the correct word) and I am on GitHub mobile, since English is my second language, I know that the last text (and maybe this one too) had many errors grammatically, but I hope it is still understandable

deepu105:
@Gabz4200 it's all perfectly fine, Thanks for taking the time to respond. English isn't my primary language as well. Grammar should be the last thing we need to worry in OSS IMO :P

deepu105:
@kRHYME7 @Gabz4200 if there is anything I can help to move this forward, I'll be happy to give it a go. I'm new to Hyprland and HyDE (Coming from Manjaro+KDE) but willing to learn and try.

kRHYME7:
I'm learning to organize some stuff with the github milestone. So contributors can see any UWSM progress.

So one step we need is to make sure we can have a proper ENV with or without UWSM.

This can be done by making fish and zsh env robust. I'm working on zsh at the moment, obviously with the breaking zsh stuff.

Will refactor fish too. 

After that UWSM can be added.

Then Remove hardcoded env vars from hyprland configuration.


---

After investigating, I learned that UWSM overrides the shell before launching the compositor.

`Shell > UWSM > compositor `


We want a functional dotfile even without UWSM.

We can do that by making the shell envs robust for zsh and fish (only this for now because that's what we maintain  on this repo.)

kRHYME7:
I'm learning to organize some stuff with the github milestone. So contributors can see any UWSM progress.

So one step we need is to make sure we can have a proper ENV with or without UWSM.

This can be done by making fish and zsh env robust. I'm working on zsh at the moment, obviously with the breaking zsh stuff.

Will refactor fish too. 

After that UWSM can be added.

Then Remove hardcoded env vars from hyprland configuration.


---

After investigating, I learned that UWSM overrides the shell before launching the compositor.

`Shell > UWSM > compositor `


We want a functional dotfile even without UWSM.

We can do that by making the shell envs robust for zsh and fish (only this for now because that's what we maintain  on this repo.)

Gabz4200:
@kRHYME7 Just realized that this exists "HYPRLAND_NO_SD_VARS=1 Disables management of variables in systemd and dbus activation environments" (more info in https://wiki.hypr.land/Configuring/Environment-variables/), so, I suppose, you can just set env_vars normally in Hyprland config (maybe in a hyprEnv.conf, so it is easy to find) and, if uwsm/env-hyprland has HYPRLAND_NO_SD_VARS, it is going to completely ignore the env on confs, giving us something that works on both.

We could, if all hyprland envs are in one hyprEnv.conf that is sourced by hyprland.conf, make a simple script that generates both it and uwsm/env-hyprland, and even put it as a pre-commit-hook just in case. Just to avoid duplicated work.

Maybe it is not what you want, I am not exactly sure about what exactly you are searching for.

kRHYME7:
Are this available in stable 0.49? or git?

Gabz4200:
Sorry about the huge delay. But it is available on stable 0.49

slapjd:
> Hi @deepu105 , as far as I can tell, it is usable as long as you stick to UWSM, but it doesn't make the branch less of a WIP.
> 
> There is just one tiny problem, it is very outdated. So you, if you merged it on your local HyDE git file (supposing you have one), you would have to, first, solve the merge conflicts that will happen on some files, then be sure to move any env declaration that still on the hyprland configuration to the uwsm env file that appeared on Configs/.config/uwsm/, since the management is different.
> 
> If you are worried about updates on the HyDE side regards envs, you can set "exec-once = uwsm finalize [ALL_ENV_VAR_NAMES_YOU_WANT]" at the end of .config/hypr/hyprland.conf, or, if you don't set any envs on userprefs, go to hyde.conf and change "SYSTEMD_SHARE_PICKER" to this same uwsm command, "uwsm finalize" does the same as the dbus-activation and systemd user import-envs, plus some uwsm things.
> 
> It may be boring if you don't know where to look at, but it works.
> 
> I think there are still some scripts on .local/lib/hyde that are not using uwsm, like hyde-launch that uses gtk-launch, but these don't break anything, it just doesn't use the full potential of uwsm.
> 
> Maybe I can make my own pull request with these fixes... But I am not confident enough to do it and don't feel I made something that is not good enough :p

Hi all, apologies for the late reply I didn't notice my github notifs.
I can rebase my branch (or merge if necessary) after I finish work. @kRHYME7 i saw you merged things into dev with merge commits. I couldn't tell if you added extra things or not to the branch, do you need me to merge to preserve any of that or can I rebase and keep the commit history clean (i.e. only actually commits that implement UWSM)

slapjd:
I'm gonna summarise my findings/opinions on environment management here, so that everyone is on the same page.

# Why isn't Hyprland's `env` good enough?
Hyprland and UWSM documentation says, essentially "if you are using uwsm, please put all of your environment inside `${XDG_CONFIG_HOME}/uwsm/env[-hyprland]`". However, what the documentation, in my opinion, doesn't really explain very well is *why* you should do this which is relevant here, for determining if we can say "actually i'm gonna put it in my compositor anyway thanks".

From what I can glean, it boils down to roughly "environment variables are not automatically given to `systemctl --user` services". i.e., if you do `systemctl --user start <thing>`, your environment of your shell is *not* transferred from your current shell. Instead, systemd has a track of its own environment variables and uses that as the environment for the service. In normal hyprland, this doesn't *really* matter that much (until it does) because you typically run things with `exec-once` or whatever and that *does* inherit the environment. Most systemd user services are not graphical so they do not need any environment variables from hyprland.

Nevertheless, Hyprland does actually export a basic set of these variables to systemd inside its `CCompositor::startCompositor()`, including important things like `$WAYLAND_DISPLAY`. This is done to allow most systemd services to at least *run*. **This is what HYPRLAND_NO_SD_VARS actually does**. If you set `$HYPRLAND_NO_SD_VARS` then Hyprland will not give systemd variables like `$WAYLAND_DISPLAY` for you, which will actually go against uwsm as uwsm somewhat expects a compositor to do that (ideally).
Additionally, environment variables made with `env` are *not* exported to systemd by Hyprland itself. This is why you need to run `systemctl --user import-environment`, which imports *all* the variables that `systemctl` can currently see to systemd for your user.

Even with this, there is one big thing that, on a normal non-UWSM setup, is not done at all which is *cleaning up* the systemd environment. As far as I can tell, if you log out of a hyprland session where you have done `systemctl --user import-environment`, the environment variables you have set are not actually cleared, and will continue to hold their value even if you login with a new session! This is what UWSM aims to prevent with its environment management. It keeps a careful track of what environment variables it has given to systemd, and will unset them for you on `uwsm stop` (or `loginctl` logout). This management makes it a bad idea to run `systemctl --user import-environment` while using UWSM, because it will not clear any extra variables that you have just passed to systemd!

This is why UWSM has a `finalize` command. `uwsm finalize` is designed for you to pass environment variables to uwsm, so that it can do the adding to systemd for you while also keeping track of them and cleaning them up afterwards (because you have told it that you have set it). In theory, nothing would prevent me from using `env=` in Hyprland and then giving *all* of the set variables to `uwsm finalize`. This would actually work perfectly fine, it's just it would be very tedious and awkward to do. `uwsm finalize` by default does not actually add any variables that are not *explicitly* given to it. The only variables it will give to systemd are the ones passed to it *and* some predefined ones in `uwsm-plugins/hyprland.sh`.

To summarize, if we use `env =` with `systemd --user import-environment` as we would normally do on non-UWSM, we would prevent UWSM from performing its environment cleanup when we log out but everything *should* run perfectly fine on the first session launched after bootup. I'm unsure if there would be any implications for UWSM's/systemd's shutdown ordering because of this.

# So if we want to maintain compatibility with non-UWSM, what can we do?
The first, somewhat obvious option is to just set the environment variables in both Hyprland and in uwsm. This will actually work perfectly fine, provided you are very careful and make absolutely sure to set the same variables in the same way in both configs. This will cause a problem though if, for some reason, a variable needs to be different between UWSM and non-UWSM, as now hyprland will overwrite it *within its environment*. This difference *should* only make a difference for processes launched without `uwsm app` in theory, but it is still odd and error-prone in my opinion and so I'd *rather* avoid overwriting variables already set if possible.

Option 1.5 would be, as mentioned by @Gabz4200, to have some sort of hyde specific config for environment that gets automatically translated to hyprland config *and* uwsm shell script. This would maintain consistency between the two at the cost of having a non-standard config for environment which users would need to learn should they want to change or add to the environment later. Not to mention that this translator script would need to be made.

The second option is to keep variables as they are and manually `uwsm finalize` *all* of them individually when running on uwsm. This is clunky and would require the same error-prone maintenance as option 1 but at least we're not writing the values twice? As far as I can tell this would work perfectly fine if it were setup correctly.

The third option is to run uwsm's env and env-hyprland before running Hyprland. They are normal shell scripts and do not require uwsm to be installed for HyDE to do this, This could be done with a special Hyde .desktop entry for non-UWSM. Hyprland would simply inherit the environment variables from the two uwsm shell scripts this way, and pass them on to its child processes perfectly fine as if you had set them with `env =`. The downside of this is that you now need a .desktop entry in `/usr/share/wayland-sessions` just for Hyde. UWSM would handle environment as it normally does and the Hyprland (UWSM) session can be used as normal. This would achieve what *I* want, which is avoiding Hyprland doing unnecessary overwriting of environment variables set by UWSM and avoiding writing out all the environment variables we need twice.

The fourth option, in theory, is to somehow get Hyprland to set environment variables *only* if not running with UWSM. I tried this before, unsuccessfully from what I remember. This would prevent Hyprland from doing the unnecessary overwrites of variables because it wouldn't try to set them if it detects UWSM. The only caveat to this option is that we would actually need some method of doing this, and it has to be *during* the config setup (otherwise some startup processes may be started with an incorrect environment).

The fifth, forbidden option is to switch entirely to UWSM and leave non-UWSM users to deal with that :stuck_out_tongue_closed_eyes:. At the moment, that's pretty much what this PR does just because that's how I set it up on my system.

And finally, you can put all these variables inside the user's shell which, under most display managers ~~except fucking gdm~~ should work fine but now you have leaked all your graphical, potentially wayland specific variables into any other session that you launch. Not to mention you still have to rewrite your environment handling many times, 1 per (non-POSIX compliant) shell which is part of the reason to use UWSM in the first place.

# Anything else?
If non-UWSM support is desired, we also need to launch apps without `uwsm app` or `app2unit` or whatever when we are not in a uwsm environment, I believe. app2unit *might* work on its own (as in, doesn't require uwsm installed in order to launch things) to still launch apps but all programs would be launched via systemd which could cause some headaches, potentially.

kRHYME7:
@slapjd Thank you! I'll probably read this a few more times. I agree with all of your points and I've reached that roadblock already. 

I already have an Idea in mind and I implemented it locally.


I'll open a separate PR soon to show the POC. I Don't mind writing multiple non posix shell. It's fine as it will be used like how we have the `.zshenv`.  I'm leaning towards that approach.

I have `~/.config/env.d` or something like that. Which declares all the shell variables.  so it can have hyde.sh, hyde.fish, gpu.sh  etc.  and be agnostic to uwsm or other shell. 

UWSM will read that directory for any .sh file and source it. 


With this approach I can also dynamically load env variables depending on my gpu. 

For now this is working for me.


---


About per application launching using uwsm we can talk about that next time.

slapjd:
> @slapjd Thank you! I'll probably read this a few more times. I agree with all of your points and I've reached that roadblock already. 
> 
> I already have an Idea in mind and I implemented it locally.
> 
> 
> I'll open a separate PR soon to show the POC. I Don't mind writing multiple non posix shell. It's fine as it will be used like how we have the `.zshenv`.  I'm leaning towards that approach.
> 
> I have `~/.config/env.d` or something like that. Which declares all the shell variables.  so it can have hyde.sh, hyde.fish, gpu.sh  etc.  and be agnostic to uwsm or other shell. 
> 
> UWSM will read that directory for any .sh file and source it. 
> 
> 
> With this approach I can also dynamically load env variables depending on my gpu. 
> 
> For now this is working for me.

If that's what you want then I can merge my stuff with that when the time comes. I will probably personally move that back out of the shell in my fork (not this pr) but that's my problem 😝 
Do you mean you'll PR into my repo or do you mean you'll PR into HyDE? Either works for me

Also, I was rebasing last night and I noticed that the new waybar.py reads environment variables by parsing `.local/state/hyde/environment` which broke waybar for me due to an empty GDK_BACKEND (which might be my fault, still testing) being explicitly set as empty (instead of unset). Why does this environment file exist? Do waybar.py or other Hyde programs not inherit the environment properly for you? Is this file used anywhere else? I would much prefer something like this didn't exist if possible because it shouldn't be necessary and is a pain to get correct with small details

kRHYME7:
~about waybar open it on  another issue please.~ Fixed

kRHYME7:
>  Do you mean you'll PR into my repo or do you mean you'll PR into HyDE? Either works for me

A separate PR for dev branch of HyDE

slapjd:
> ~about waybar open it on  another issue please.~ Fixed

Thanks lol. Do you think it's a good idea to get rid of that environment file altogether (as in, don't generate it) along with the new environment system you're cooking?

> >  Do you mean you'll PR into my repo or do you mean you'll PR into HyDE? Either works for me
> 
> 
> 
> A separate PR for dev branch of HyDE 

Got it. Cheers

slapjd:
Rebased onto dev. Should all be working now

kRHYME7:
> Do you think it's a good idea to get rid of that environment file altogether (as in, don't generate it) along with the new environment

That exists to look up for hyprland variables being set. It was very helpful when UWSM was not yet popular at that time.

kRHYME7:
Hi I will personally add the UWSM support soon after hyprlang hits the extra arch  repo.

This should do the trick
https://github.com/hyprwm/hyprlang/issues/52#issuecomment-3120075593

slapjd:
> Hi I will personally add the UWSM support soon after hyprlang hits the extra arch  repo.
> 
> This should do the trick
> https://github.com/hyprwm/hyprlang/issues/52#issuecomment-3120075593
> 
> 

That looks like it solves our problems, yes. Is the plan to import environment if some UWSM variable is unset?

kRHYME7:
Yes something like that.

Also  Looks like no performance issues even adding multiple conditional environment variables. I've been testing this for a week now.

PR #569, opened 2025-05-04
add best fzf aliases ever

prime-run:
## Description

This PR introduces three functions to `.zshenv` taking advantage of `fzf`, aiming for better and faster Terminal navigation.

The mentioned aliased functions are as follows:
```bash
- f_efc='_fuzzy_edit_search_file_content'       # [fz]f [e]dit  [f]ile, search by [c]ontent
- fcd='_fuzzy_open_directory'             # [fz]f [o]pen [d]irectory
- f_ef='_fuzzy_edit_search_file'          # [fz]f [e]dit [f]ile
```

This approach takes advantage of `fzf` by directly inputting `query` which benefits us by:
- a) generally faster navigation
- b) triggering fzf filters at once

#### Some extra features:
1- excluding unwanted search results in `fod` function
2- null `$EDITOR` fallback
3- adjustable `max_depth=5` for fuzzy finding
4- etc...

## Type of change

- [x] **Documentation update** (non-breaking change; modified files are limited to the documentations)

## Checklist

### `fcd`

https://github.com/user-attachments/assets/64a46d47-186e-4b20-b8f1-5b1736487e4b

instead of doing : `cd ~/Projects/rust/wallrust/src`
we just `fuzzy` it via `fcd` 

video demonstrates 2 fuzzy "searchings" for the target dir!

------

### `f_efc`

https://github.com/user-attachments/assets/f92715a3-fb31-486e-a070-94dbf15bfe13

we fuzzy find files containing target keyword (in this case `save` and `os`) in the directory and selecting it opens the file in $EDITOR

## Additional context

- This features can elevated with tools like `bat` and `ripgrep`
- same aliases for `fish` shell will be pushed in a different PR

### DIFF
--- a/Configs/.zshenv
+++ b/Configs/.zshenv
@@ -141,7 +141,7 @@ _fuzzy_open_directory() {
     fi
 
     #type -d
-    selected_dir=$(find . -maxdepth max_depth -type d \( -name .git -o -name node_modules -o -name .venv -o -name target \) -prune -o -print 2>/dev/null | fzf "${fzf_options[@]}")
+    selected_dir=$(find . -maxdepth $max_depth -type d \( -name .git -o -name node_modules -o -name .venv -o -name target \) -prune -o -print 2>/dev/null | fzf "${fzf_options[@]}")
 
     if [[ -n "$selected_dir" && -d "$selected_dir" ]]; then
         cd "$selected_dir" || return 1 #  if cd fails
@@ -161,7 +161,7 @@ _fuzzy_edit_search_file() {
     fi
 
     # -type f: only find files
-    selected_file=$(find . -maxdepth max_depth -type f 2>/dev/null | fzf "${fzf_options[@]}")
+    selected_file=$(find . -maxdepth $max_depth -type f 2>/dev/null | fzf "${fzf_options[@]}")
 
     if [[ -n "$selected_file" && -f "$selected_file" ]]; then
         if command -v "$EDITOR" &>/dev/null; then
--- a/Configs/.zshenv
+++ b/Configs/.zshenv
@@ -129,6 +129,60 @@ function _load_omz_on_init() {
     fi
 }
 
+# best fzf aliases ever
+_fuzzy_open_directory() {
+    local initial_query="$1"
+    local selected_dir
+    local fzf_options=('--preview=ls -p {}' '--preview-window=right:60%')
+    local max_depth=5
+
+    if [[ -n "$initial_query" ]]; then
+        fzf_options+=("--query=$initial_query")
+    fi
+
+    #type -d
+    selected_dir=$(find . -maxdepth max_depth -type d \( -name .git -o -name node_modules -o -name .venv -o -name target \) -prune -o -print 2>/dev/null | fzf "${fzf_options[@]}")
+
+    if [[ -n "$selected_dir" && -d "$selected_dir" ]]; then
+        cd "$selected_dir" || return 1 #  if cd fails
+    else
+        return 1
+    fi
+}
+
+_fuzzy_edit_search_file() {
+    local initial_query="$1"
+    local selected_file
+    local fzf_options=()
+    local max_depth=5
+
+    if [[ -n "$initial_query" ]]; then
+        fzf_options+=("--query=$initial_query")
+    fi
+
+    # -type f: only find files
+    selected_file=$(find . -maxdepth max_depth -type f 2>/dev/null | fzf "${fzf_options[@]}")
+
+    if [[ -n "$selected_file" && -f "$selected_file" ]]; then
+        $EDITOR "$selected_file"
+        #FIX: if ! $EDITOR case
+    else
+        return 1
+    fi
+}
+
+_fuzzy_edit_search_file_content() {
+    # [f]uzzy [e]dit  [s]earch [f]ile [c]ontent
+    local selected_file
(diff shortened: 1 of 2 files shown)


Configs/.zshenv
```diff
@@ -143,6 +207,9 @@ function _load_post_init() {
     # Initiate fzf
     if command -v fzf &>/dev/null; then
         eval "$(fzf --zsh)"
+        alias f_efc='_fuzzy_edit_search_file_content' # [fz]f [e]dit  [f]ile, search by [c]ontent
+        alias fcd='_fuzzy_open_directory'             # [fz]f [o]pen [d]irectory
+        alias f_ef='_fuzzy_edit_search_file'          # [fz]f [e]dit [f]ile
```
kRHYME7: IMO it's better to have a conventional namespace for fzf.

Example:

We use `ff` as prefix (I guess it's not straining than flat_`

ffcd - Fuzzy Finder Change Directory
ffe - Fuzzy Finder Edit (edit and search keywords can be omitted as Fuzzy Finder already indicates a Search and in linux everything is a file)
ffec - Fuzzy Finder Edit Content (Consistent according to above with [c] being treated as a flag. Edit Content)

kRHYME7 (approved):

kRHYME7:
can we just merge `edit ` and the ` edit content` with preview ? any reasons not having   a preview?

kRHYME7:
@prime-run  Might be too much but check this out first. 

Also, the older implementation seems useless if you have files that are deeply nested. so I explicity add find command on every level. Though, it makes the function long. I guess this can be refactored. 

### Preview

https://github.com/user-attachments/assets/5e342ad2-146d-49c6-9ab2-b2fe9fb9085d

Configs/.zshenv
```diff
@@ -129,6 +129,246 @@ function _load_omz_on_init() {
...
+    local search_term=""
+
+    # Parse command arguments
+    while [[ $# -gt 0 ]]; do
+        case "$1" in
+        --content)
+            content_search=true
+            shift
+            ;;
+        --regex)
```
prime-run: this can go so wrong on lower end machines!

prime-run: >  useless if you have files that are deeply nested

could please elaborate with an example.

prime-run:
and about `_fuzzy_change_directory`, I think it's killing the intent of the ailias! (I think it should be a different ailias)


like in the screen shot below, the goal is to cd to `/home/<username>/.config/hyde/themes/Decay Green` by fuzzy searching `hyde/decay`
am i missing something in the new `_fuzzy_change_directory` implementation?

250505_10h54m19s_screenshot

Configs/.zshenv
kRHYME7: I can't search .zshrc because there is zshrc in .config/cfg_backups/.zshrc which always comes first.

also better to load shallower files than deep.

kRHYME7: what could cause a bootleneck in here? regex? or multiple find command?

kRHYME7:
Ah yeah. I see... how about the other functionality?

I'll revert the ffod later as I understand the intent.

kRHYME7:
Owww Cool!

Configs/.zshenv
prime-run: > I can't search .zshrc 

unless you mean `.zshrc\` directory, I'd call it a feature not a bug😄 (the `-d` flag in find command)  

------ 

> what could cause a bootleneck

I would argue the whole `while` loop! 
you can monitor your memory allocation to confirm!

Configs/.zshenv
prime-run: > better to have a conventional namespace for fzf.

I AGREE! 
I'll paypal you $0.42 if you take over the namings 😄😄

prime-run:
@kRHYME7 

> ffcd - Fuzzy Finder Change Directory
ffe - Fuzzy Finder Edit (edit and search keywords can be omitted as Fuzzy Finder already indicates a Search and in linux everything is a file)
ffec - Fuzzy Finder Edit Content (Consistent according to above with [c] being treated as a flag. Edit Content)

is this final ?
I can apply them and nuke `new implementations` for now.
how does that sound ?

Configs/.zshenv
kRHYME7: I currently have 0.0011 so that would be a lot 😂

prime-run: i like how my PRs literally turn into a discor chat ! 😄😄

kRHYME7:
> maybe we can move our fzf stuff to a seprate file and built on top of that in future if that fits your vision

We can add it as a plugin on our plugin installation for zsh or make it a core script and make it POSIX as much possible. 

---

> is this final ?

This
- ffe
- ffec
- ffod

I don't mind nuking the `_fuzzy_search_directory` (former `_fuzzy_change_directory`) and use `ffod ` for that. 

But please add fzf args to make it like a inline pop menu not a fullscreen (So it feels natural drop down ).

image

```
  fzf_options+=(
        --height "80%"
        --layout=reverse
        --preview-window
        right:60% --cycle
    )

```


---

About `ffec`  I can't really navigate it well specially `.config/` always comes first. and fzf just  takes eternity  searching all files. For now this will do but gotta find a way to fuzzy search  plus prioritize shallow files.

kRHYME7:
Cool, ready for merging?

prime-run:
> We can add it as a plugin on our plugin installation for zsh or make it a core script and make it POSIX as much possible.

so let me pull of the exact same thing (`ffe`, `ffec`, `ffcd`) for `fish`

cool ?

PR #641, opened 2025-05-19
feat: Auto execute only in kitty

Oidaho:
# Pull Request

## Description

This commit adds a condition to the shell configuration that guarantees execution of commands that start with the terminal only in the kitty terminal emulator.

This change is primarily aimed at eliminating the `fastfetch` or `pokego` call in terminal emulators such as the Visual Studio Code terminal.

## Type of change

## Checklist

## Screenshots

vsc

kitty

### DIFF
--- a/Configs/.user.zsh
+++ b/Configs/.user.zsh
@@ -2,12 +2,17 @@
 # Commands to execute on startup (before the prompt is shown)
 # This is a good place to load graphic/ascii art, display system information, etc.
 
-if command -v pokego >/dev/null; then
-    pokego --no-title -r 1,3,6
-elif command -v pokemon-colorscripts >/dev/null; then
-    pokemon-colorscripts --no-title -r 1,3,6
-elif command -v fastfetch >/dev/null; then
-    fastfetch --logo-type kitty
+# The commands will be executed only at the start of Kitty.
+if [[ -n "$KITTY_PID" ]]; then
+
+    if command -v pokego >/dev/null; then
+        pokego --no-title -r 1,3,6
+    elif command -v pokemon-colorscripts >/dev/null; then
+        pokemon-colorscripts --no-title -r 1,3,6
+    elif command -v fastfetch >/dev/null; then
+        fastfetch --logo-type kitty
+    fi
+
 fi
 
 # fastfetch.sh


kRHYME7:
We also want to show images on other emulators that support kitty protocol (supports images). How can we possibly do that?

Oidaho:
> We also want to show images on other emulators that support kitty protocol (supports images). How can we possibly do that?

I think we can use something like that:
```bash
supports_kitty_graphics() {
    # kitty
    if [[ -n "$KITTY_PID" ]]; then
        return 0
    fi

    # Alacritty
    if ps -o comm= -p "$PPID" | grep -q "alacritty"; then
        return 0
    fi

    if [[ "$TERM_PROGRAM" == "WezTerm" ]]; then
       return 0; 
    fi

    return 1
}

```

However, this solution doesn't seem very elegant to me. The problem here lies in the fact that emulators that support the kitty protocol, for the most part, do not create a variable, by analogy with `$KITTY_PID`.

In this case, I decided to build on the fact that Kitty is pre-installed and officially supported by HyDE.

rubiin (changes requested):
Besides kitty, we should also consider terms like wezterm, alacritty as discussed on the replies

kRHYME7:
@Oidaho  Thank you! I already resolve this in https://github.com/HyDE-Project/HyDE/commit/2b295eb148c7a4444934303c772ab1a9062439ff

Oidaho:
> @Oidaho Thank you! I already resolve this in [2b295eb](https://github.com/HyDE-Project/HyDE/commit/2b295eb148c7a4444934303c772ab1a9062439ff)

You're welcome!
It's a pity that I didn't have enough time to finish my PR. :(

kRHYME7:
BTW I just made a function and added some Known Terminal that renders an image. so 

if do_render image;then command for image;fi

There's no conventional way to do it so let's just do it our way for now.

Oidaho:
> There's no conventional way to do it so let's just do it our way for now.

Actually, I tried to implement it through a different approach.
My idea was not just to determine the tty, but to check whether the tty supports the kitty graphics protocol.

If I come up with something, I'll recreate this PR.

PR #672, opened 2025-05-25
Fix: togglefloating keybind

Hello-Meow:
# Pull Request

## Description

This PR fixes the togglefloating keybind behavior. Currently, the keybind will resize the window to 95% every time it is used. This is fine when the window is floating, but it causes a tiled window to blow up and push all other tiles to the edge of the workspace.

## Type of change

## Checklist

## Additional context

Hyprland's default togglefloating behavoir is a bit quirky. The initial size of floating windows is set to the window's size when it is created. When a tiled window is resized due to other windows being created/moved/etc, the stored size of the floating window does not change, so a window that is made to float for the first time can seem to suddenly blow up.

This was changed here (https://github.com/HyDE-Project/HyDE/pull/543), by giving a floating window a consistent size of 95%, which caused the issue this PR attempts to fix.


kRHYME7:
@prime-run Hello! This is an example of why I want to add an ipc. 

Basically it reads Hyprland's event.

So for cleaner keybindings.conf we can just use ` ..., togglefloating`  dispatch.

Now when hyprland receives that signal we can "react" to that by handling the resize etc. 

IMO this is better than making too many wrapper scripts just for basic Hyprland stuff.


This way the abstraction is well organized and layered.

1. dispatch Hyprland stuff
2. Use IPC so HyDE can add its own behavior that is not in Hyprland config.

User can then tweak the the reactions.

prime-run:
@kRHYME7 hi! sorry for my absence in the last few days! My city was hit by a tornado, and the internet was pretty spotty for a few days. It seems to be fixed now! I've added alot to `hyde-ipc` and I'm doing will do final checks I'll push the updates then get back to this issue.

------
> This way the abstraction is well organized and layered.

I should mention : 
and also
- Uses hyprland sockets and shared states directly so it's the fastest way of dispatching,
- it's handled asynchronously so doesn't actually block anything

PR #693, opened 2025-05-28
feat: use `duf` as alias for `df`

zeerayne:
# Pull Request

## Description

`duf` does the same as `df`, but with pretty output. Simple aliases would not work because `df` and `duf` have different options. Functions for zsh and fish drop all `df` arguments except of path, so `duf` is called without arguments or with path only.

## Type of change

## Checklist

### Before

image
image

### After

image
image


TowhidMolla-Dev:
That's dope man

PR #795, opened 2025-06-12
fix: include nixos wrapped program to prevent infinite reloads

richen604:
# Pull Request

## Description

fixes infinite reload of the waybar python script because nixos is special and has special named programs :sweat_smile: 

## Type of change

## Checklist

## Screenshots

image

## Additional context

on a sidenote i have a sed script already that replaces `killall waybar` with `.waybar-wrapped` globally, weirdly `pkill waybar` somehow works

### DIFF
--- a/Configs/.local/lib/hyde/waybar.py
+++ b/Configs/.local/lib/hyde/waybar.py
@@ -1403,7 +1403,7 @@ def watch_waybar():
                 time.sleep(2)
                 continue
 
-            result = subprocess.run(["ps", "-C", "waybar"], capture_output=True)
+            result = subprocess.run(["ps", "-C", "waybar,.waybar-wrapped"], capture_output=True)
             if result.returncode != 0:
                 run_waybar_command("killall waybar; waybar & disown")
                 logger.debug("Waybar restarted")


kRHYME7:
What happens if we use pkill?  do we still need the waybar-wrap ?

richen604:
nope pkill works directly with `waybar`, no idea why

kRHYME7:
Let's just use pkill then.

richen604:
would you want something like that in a separate MR? I'm fine with this as is for now

looking into it `pkill waybar` would be the same as `killall "*waybar*"` and match all processes with waybar in the name. might not be ideal

kRHYME7:
I find `.waybar-wrapped` confusing as per non nix so  i am trying to compensate in using pkill. 

I think we can do exact matching  for pkill

```sh
pkill -x waybar
``` 
should do the trick

richen604:
if you were to do something explicit it would have to match both processes, eg `killall -r "waybar|\.waybar-wrapped"`
i try to get around this using sed because its kitty, dunst, and waybar that only have wrapped versions, so let me worry about that confusion in hydenix

`ps -C waybar,.waybar-wrapped` will match both processes, if this can be merged i have less sed commands to maintain

kRHYME7:
I'm still confused 😂 But ok 👌

PR #797, opened 2025-06-12
feat: wallbash window border colors

zeerayne:
# Pull Request

## Description

This PR offers wallbash-dependant window border colors to match wallpapers and bring more style. But I'm not sure about achitectural way to to this properly. Review, comment and ideas are required.

## Type of change

## Checklist

## Screenshots
image

image

image

image

image

### DIFF
--- a/Configs/.local/lib/hyde/waybar.py
+++ b/Configs/.local/lib/hyde/waybar.py
@@ -1403,7 +1403,7 @@ def watch_waybar():
                 time.sleep(2)
                 continue
 
-            result = subprocess.run(["ps", "-C", "waybar"], capture_output=True)
+            result = subprocess.run(["ps", "-C", "waybar,.waybar-wrapped"], capture_output=True)
             if result.returncode != 0:
                 run_waybar_command("killall waybar; waybar & disown")
                 logger.debug("Waybar restarted")


zeerayne:
@kRHYME7 Please check this PR, your opinion is required here

kRHYME7:
So when should these borders take effect? I think wallbash mode already has its borders. Theme mode should have its own too. We also have a  fallback  for groups so maybe add fallback for the border too?

zeerayne:
The idea was to use these borders "always", like it is done for groupbar backgrounds:
```hyprlang
group:groupbar {
    col.inactive = rgba($wallbash_pry3ee)
    col.active = rgba($wallbash_pry1ee)
    col.locked_active = rgba($wallbash_pry2ee)
    col.locked_inactive = rgba($wallbash_pry4ee)
}
```

kRHYME7:
always? when? how to activate it?

hmm, one can just add it in userprefs though, so I don't understand. In wallbash mode, borders are using wallbash so, I think this one is redundant? Or I didn't see the greater picture of the use case of this one?


We already have global group bar, which of course can be overwritten by a theme (though mostly they don't care )
https://github.com/HyDE-Project/HyDE/blob/b2e2c21c97668e417322cbb56ce57b7f43b0d9e0/Configs/.local/share/hyde/hyprland.conf#L261C1-L267C2

We can add a global border below that (only take effect if themes do not have border colors)

---


> hmm, one can just add it in userprefs though, 

> always? when? how to activate it?

Replying to my self, what can I think of, given the circumstance we can still add that borders.conf  and it is on the user to do something like this 

```
source = ./themes/borders.conf

```

zeerayne:
> Replying to my self, what can I think of, given the circumstance we can still add that borders.conf and it is on the user to do something like this

You mean empty config which can be configured by user?

kRHYME7:
It can be in userprefs.conf and should just work overriding everything including themes.

zeerayne:
Got it, so I think this PR should be closed

PR #819, opened 2025-06-16
fix: Improve the _fuzzy alias

frankjuniorr:
# Pull Request

## Description

Hey @kRHYME7 and @prime-run , I found this [PR](https://github.com/HyDE-Project/HyDE/pull/569) fantastic. It added these `_fuzzy*` aliases, and they’re honestly the best fzf aliases ever. Inspired by the original PR’s description, in the **Additional context** section, I decided to give my two cents by following the suggestion.

I tried to improve the 3 aliases by adding some tools:
- `bat`, to replace `cat`
- `ripgrep` to replace `grep`
- `fd` to replace `find`
- `lsd` to replace `ls`

Since they are dependencies, I added these 3 tools to the `pkg_core.lst` file.

## Type of change

## Checklist


kRHYME7:
I want a fallback. I don't want to add those in the core packages.

Hyprdots/HyDE always tries to shred the dependencies to minimum. but here we go adding more. Let's stay lightweight and functional. 


I'll try to edit this PR to preserve gnu core utils.

prime-run:
@frankjuniorr thank you. 

I think you can find our original conversation in that PR that we opt out of these tools. altho there is nothing wrong with the `glibc` tools, I personally use the modern ones same as the ones you added in this PR. But I believe the general policy of the hyde project is to keep core packages as minimal as possible.

Note: If this PR is being considered, plz make sure to include them in `fish` config and corresponding functions as well

frankjuniorr:
Ok, got it, no problem. Makes sense.
In that case, I guess it's better to disregard this PR, right?

Configs/.zshenv
```diff
@@ -181,7 +189,13 @@ _fuzzy_edit_search_file() {
     fi
 
     # -type f: only find files
-    selected_file=$(find . -maxdepth $max_depth -type f 2>/dev/null | fzf "${fzf_options[@]}")
+    selected_file=$(fd --type f --max-depth $max_depth \
+      --exclude .git \
```
prime-run, line 193: `fd` has a builtin flag for respecting `.gitignore` in any depth. there is no need for manual exclusion.

Configs/.zshenv
```diff
@@ -133,7 +133,7 @@ function _load_omz_on_init() {
 _fuzzy_change_directory() {
     local initial_query="$1"
     local selected_dir
-    local fzf_options=('--preview=ls -p {}' '--preview-window=right:60%')
+    local fzf_options=('--preview=lsd {}' '--preview-window=right:60%')
```
prime-run, line 136: does `fzf` parse this ? I think 
```bash
lsd -1
```
is more "fuzzable"!

rubiin:
@kRHYME7 maybe we can add a step where the install scripts asks to install these and based on user input the packages are installed. This gives more freedom and has best of both worlds . This would mean we need to setup a config similar to $IDLE like $FINDER=fd/find

kRHYME7:
@rubiin  Yup, but I guess it is time for this #826  and then by that it is cleaner to install optional stuff.

kRHYME7:
This  changes can be added as `./functions` or `./conf.d` following this PR https://github.com/HyDE-Project/HyDE/pull/858

also please move those extra packages to `pkg_extra.lst` not ` pkg_core.lst`

frankjuniorr:
@kRHYME7  I believe this PR should be closed because its original purpose has been completely replaced by this PR here (#858). Btw, your solution in this PR (#858 ) looks really good, I’ll test it in my environment, but I think the one in this thread can be closed as there’s not much left to do with it.

PR #865, opened 2025-06-24
Mediaplayer

heeeeeeeeeeh:
# Pull Request

## Description

-  24e1ee3 feat(mediaplayer): Added waybar module for mediaplayer
- 49d9a32 feat(mediaplayer): modified mediaplayer to support multiple players. I have multiple players (for offline and online use)
- 266a7ad feat(mediaplayer): Add env option for players. so that it is easier to set players
- c1aace0 fix(mediaplayer): get first player that is player(I meant playing) if there are none then get first player in list. User is more likely to want to be shown the playing then the first option they gave for player.
- faf674f feat(mediaplayer): Added media controls (next/previous/playpause). used signals bc I was lazy and it seem the easiest way to do it
- 8694197 fix(mediaplayer): escaped & in song metadata. having a & in metadata would cause an error
- 736edb1 Made text only show main artist Tooltip shows all. I think the title is more important.
- 053c9a5  fix(mediaplayer.py): redo algorithm that proportions (forgot to add text). I did this because it makes sure that if either artist or title has some extra space it passes it to the other if needed
- feat(mediaplayer.py): Added env option for artist weight. makes mediaplayer more versatile 
- 39943f6 fix(mediaplayer): moved all players output to tooltip except for current. this makes sure that waybar only  shows the current player. if write_output is called for every one that would be unnecessary because only  one player's metadata will show and it wouldn't be the current player
- ec5e588 - feat(mediaplayer): added player name to tooltip. to know that is playing what. Imma add option for the color later. Maybe there should be an option for custom icons for  player in the text field and a default as fallback 

## Type of change

- [x ] **Bug fix** (non-breaking change which fixes an issue)

## Checklist

## Screenshots

(if appropriate)

## Questions
This is my first ever pull request so I'm a bit confused. 

- I added some env options to mediaplayer but i wasn't sure how to update the document to reflect it and add it the options to the schema file
- It the contributing docs it says to make sure all the tests pass but Im not sure how to run them.
- I not sure how I've should have bundle the commits. Is like this fine, or should I have separated them into different PRs
- Where is the changelog?
- I wanted to add the progress bar to text field too but idk if that a good idea

### DIFF
--- a/Configs/.local/lib/hyde/mediaplayer.py
+++ b/Configs/.local/lib/hyde/mediaplayer.py
@@ -108,10 +108,10 @@ def write_output(track, artist, playing, player, tooltip_text):
     logger.info("Writing output")
 
     output_data = {
-        "text": format_artist_track(artist, track, playing, max_length_module),
+        "text": escape(format_artist_track(artist, track, playing, max_length_module)),
         "class": "custom-" + player.props.player_name,
         "alt": player.props.player_name,
-        "tooltip": tooltip_text,
+        "tooltip": escape(tooltip_text),
     }
 
     sys.stdout.write(json.dumps(output_data) + "\n")
@@ -375,6 +375,10 @@ def set_player(manager, player):
 def control_music(sig, frame, action):
     if currentplayer:
         getattr(currentplayer, action)()
+
+
+def escape(string):
+    return string.replace("&", "&amp;")
 
 
 if __name__ == "__main__":
--- a/Configs/.local/lib/hyde/mediaplayer.py
+++ b/Configs/.local/lib/hyde/mediaplayer.py
@@ -264,7 +264,7 @@ def parse_arguments():
 
 def main():
     global prefix_playing, prefix_paused, max_length_module, standby_text, artist_track_separator
-    global artist_color, track_color, progress_color, empty_color, time_color
+    global artist_color, artist_weight, track_color, progress_color, empty_color, time_color
 
     # Load environment variables from your config file:
     config_file = os.path.join(xdg_state_home(), "hyde", "config")
@@ -298,6 +298,7 @@ def main():
     time_color = os.getenv(
         "MEDIAPLAYER_TOOLTIP_TIME_COLOR", "#" + os.getenv("dcol_txt1", "FFFFFF")
     )
+    artist_weight = os.getenv("MEDIAPLAYER_ARTIST_WEIGHT", 0.65)
     players = os.getenv("MEDIAPLAYER_PLAYERS", None)
     if players:
         players = players.split(",")
--- a/Configs/.local/lib/hyde/mediaplayer.py
+++ b/Configs/.local/lib/hyde/mediaplayer.py
@@ -90,9 +90,18 @@ def format_artist_track(artist, track, playing, max_length):
         if full_length > max_length:
             # proportion how to share max length between track and artist
             artist_weight = 0.65
-            track_weight = 1 - artist_weight
-            artist_limit = int(max_length * artist_weight)
-            track_limit = int(max_length * track_weight)
+            artist_limit = min(int(max_length * artist_weight), len(artist))
+            a_gain = max(0, artist_weight - (artist_limit / max_length))
+            track_weight = 1 - artist_weight + a_gain
+            track_limit = min(int(max_length * track_weight), len(track))
+            t_gain = max(0, track_weight - (track_limit / max_length))
+
+            if a_gain == 0 and t_gain > 0:
+                gain = int(max_length * t_gain)
+                artist_limit = artist_limit + gain
+            elif a_gain > 0 and t_gain == 0:
+                gain = int(max_length * t_gain)
+                artist_limit = artist_limit + gain
 
             if len(artist) != len(artist[:artist_limit]):
                 artist = artist[:artist_limit].rstrip() + "…"
--- a/Configs/.local/lib/hyde/mediaplayer.py
+++ b/Configs/.local/lib/hyde/mediaplayer.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python3
 import os
+from re import split
 import gi
 
 gi.require_version("Playerctl", "2.0")
@@ -285,6 +286,9 @@ def main():
     time_color = os.getenv(
         "MEDIAPLAYER_TOOLTIP_TIME_COLOR", "#" + os.getenv("dcol_txt1", "FFFFFF")
     )
+    players = os.getenv("MEDIAPLAYER_PLAYERS", None)
+    if players:
+        players = players.split(",")
 
     arguments = parse_arguments()
     player_found = False
@@ -299,8 +303,10 @@ def main():
     logger.debug("Arguments received {}".format(vars(arguments)))
 
     manager = Playerctl.PlayerManager()
-    if not arguments.players:
-        arguments.players = [name.name for name in manager.props.player_names]
+    if not arguments.players and not players:
+        players = [name.name for name in manager.props.player_names]
+    elif arguments.players:
+        players = arguments.players
     loop = GLib.MainLoop()
 
     manager.connect(
@@ -312,16 +318,16 @@ def main():
     signal.signal(signal.SIGTERM, signal_handler)
     signal.signal(signal.SIGPIPE, signal.SIG_DFL)
 
-    found = [None] * len(arguments.players)
+    found = [None] * len(players)
     for player in manager.props.player_names:
-        if arguments.players is not None and player.name not in arguments.players:
+        if players is not None and player.name not in players:
             logger.debug(
                 "{player} is not the filtered player, skipping it".format(
                     player=player.name
                 )
             )
             continue
-        found[arguments.players.index(player.name)] = player
+        found[players.index(player.name)] = player
         init_player(manager, player)
         player_found = True
 
--- a/Configs/.local/lib/hyde/mediaplayer.py
+++ b/Configs/.local/lib/hyde/mediaplayer.py
@@ -86,6 +86,7 @@ def format_artist_track(artist, track, playing, max_length):
             track = track[:max_length].rstrip() + "…"
         output_text = f"{prefix}{prefix_separator}<b>{track}</b>"
     elif track and artist:
+        artist = artist.split(",")[0].split("&")[0].strip()
         if full_length > max_length:
             # proportion how to share max length between track and artist
             artist_weight = 0.65
--- a/Configs/.local/lib/hyde/mediaplayer.py
+++ b/Configs/.local/lib/hyde/mediaplayer.py
@@ -24,6 +24,7 @@ logger = logger.get_logger()
 # for each player.  Key = player_name
 #
 players_data = {}
+currentplayer = None
 
 
 def load_env_file(filepath: str) -> None:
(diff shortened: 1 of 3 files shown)


kRHYME7:
fix you branch. Try making the branch against dev.

kRHYME7:
Maybe recreate  for cleaner commit history

heeeeeeeeeeh:
> fix you branch. Try making the branch against dev.

That my  bad, though maybe that should be said explicitly in CONTRIBUTING.md. I was following the doc line by line so I didn't think to switch to dev before creating branch

> Maybe recreate for cleaner commit history

What does cleaner commit history mean? Do I make a new branch and pull request?

kRHYME7:
> What does cleaner commit history mean? Do I make a new branch and pull request?

git switch dev
git checkout -b new_cool_stuff


> That my bad, though maybe that should be said explicitly in CONTRIBUTING.md. I was following the doc line by line so I didn't think to switch to dev before creating branch


some PR only depends on what should be added or be fixed, etc. Some commits  regress the master branch therefore we should patch it.

Newer features can/should be  in dev branch.

And a patch for the new_feature branch will be against that new_feature branch that is against dev branch.


Feel free to make a PR 😜. 

Here's how I map it in my head. 

image

ow ok, just got in my laptop and the commit should be clean now

Configs/.local/share/waybar/modules/custom-mediaplayer.jsonc
```diff
@@ -0,0 +1,10 @@
+{
+  "custom/mediaplayer": {
+    "exec": "hyde-shell mediaplayer.py",
+    "return-type": "json",
+    "format": "{}",
+    "on-click": "pkill -SIGUSR2 -f 'python3 /home/silve/.config/waybar/waybar-mediaplayer/src/mediaplayer monitor'",
```
kRHYME7: want to restart the mediaplayer? 

also hardcoded

heeeeeeeeeeh:
What should I do now. I pretty sure I need to update docs to list the new options in mediaplayer, but not sure how

heeeeeeeeeeh:
is it this file .local/share/hyde/schema/config.toml?

kRHYME7:
yes that file.

then use the script to generate other stuff. But it's fine if I update that for you.

kRHYME7 (changes requested):

Configs/.local/share/waybar/modules/custom-mediaplayer.jsonc
heeeeeeeeeeh: oops I just realized the wrong file location, but no its for the media  controls.
```
    signal.signal(signal.SIGUSR1, lambda *args: control_music(*args, "previous"))
    signal.signal(signal.SIGUSR2, lambda *args: control_music(*args, "next"))
    signal.signal(signal.SIGRTMIN, lambda *args: control_music(*args, "play_pause"))

def control_music(sig, frame, action):
    if currentplayer:
        getattr(currentplayer, action)()
```

heeeeeeeeeeh: SIGUSR1 is for previous track
SIGUSR2 is for next track
SIGRTMIN is for play_pause

kRHYME7: Cool but I  that's limiting can we use 

pkill +RTMIN INT ?

INT is like the SIGRTMIN signal in waybar

heeeeeeeeeeh: so I use numbers after RTMIN for the signals?

kRHYME7: Yup that way we can extend that. Do signal lib handle this in python too?

heeeeeeeeeeh: I tested the signals by running the python program and not through waybar so that is how it was wrong

heeeeeeeeeeh: > Yup that way we can extend that. Do signal lib handle this in python too?

let me check

heeeeeeeeeeh: this is how it is rn. is there an env variable where the hyde-shell scripts are at?
{
  "custom/mediaplayer": {
    "exec": "hyde-shell mediaplayer.py",
    "return-type": "json",
    "format": "{}",
    "on-click": "pkill +RTMIN 1 -f 'python3 /home/silve/.config/lib/hyde/mediaplayer'",
    "on-scroll-up": "pkill +RTMIN 2 -f 'python3 /home/silve/.config/lib/hyde/mediaplayer'",
    "on-scroll-down": "pkill +RTMIN 3 -f 'python3 /home/silve/.config/lib/hyde/mediaplayer'",
  },
}

heeeeeeeeeeh: > Yup that way we can extend that. Do signal lib handle this in python too?

Yes it does

kRHYME7: There's none.

Should I make the `hyde-shell` command output the realpath so it can be targeted by the signal dynamically?

heeeeeeeeeeh: ouu that a good idea

heeeeeeeeeeh: I just realized that waybar opens multiple processes for multiple monitors. Is there a way to tell it to use one process for both monitors? if not we can get the pid of one process and signal one process

heeeeeeeeeeh: it does the task twice if its not fixed

kRHYME7: > I just realized that waybar opens multiple processes for multiple monitors. Is there a way to tell it to use one process for both monitors? if not we can get the pid of one process and signal one process

Yeah, waybar can't do singletons for custom processes. My guess is they should be built in with the scripts.

kRHYME7: I can only think of exposing a socket.

heeeeeeeeeeh: okay I can do that tomorrow. Rn is time for me to go to sleep

kRHYME7: No problem! Take your Time

Configs/.local/lib/hyde/mediaplayer.py
```diff
@@ -240,14 +262,14 @@ def parse_arguments():
     )
 
     # Define for which player we're listening
-    parser.add_argument("--player", help="Specify the player to listen to.")
+    parser.add_argument("--players", nargs="*", type=str)
```
zeerayne, line 269: this will break `custom-spotify.jsonc` actions:
```json
"exec": "hyde-shell mediaplayer --player spotify",
```

heeeeeeeeeeh: I can put back the argument and it will only look for one player

zeerayne: ... or you can adapt existing module for new param name

Configs/.local/share/waybar/modules/custom-mediaplayer.jsonc
heeeeeeeeeeh: This is kinda hacky, but can we do:
```
kill -RTMIN $(pgrep -f /home/silve/.local/lib/hyde/mediaplayer.py | head -n 1)
```
both process should be tracking the same player.

Configs/.local/lib/hyde/mediaplayer.py
heeeeeeeeeeh: that what I meant

heeeeeeeeeeh:
How do I remove [e6e86e3](https://github.com/HyDE-Project/HyDE/pull/865/commits/e6e86e332b85acac8cb8c112da5548ae76c14049) didn't mean to push that

kRHYME7:
@heeeeeeeeeeh  

Hi! Instead of parsing the whole path's PID how abot  we cache the PID 

1. on launch save its own PID to a structured file 
example. 

`${XDG_RUNTIME_DIR}/hyde/mediaplayer_{player_name}.pid`

- Note that only create a cache when the `args` are not related to control example `mediaplayer.py` and `mediaplayer.py --player ""` should create a cache but `mediaplayer.py --pause ` shouldn't

- This can have multiple pid in it. For simplicity (for now), we can store old pid too even if it is not used.  I think pid won't ever repeat suring the whol session. and `XDG_RUNTIME_DIR` will reset during relogins/reboots.

2. Now we can do toggles using that meta file 
    We can self inject the SIGNAL, so we can have args like `mediaplalyer.py --pause --player "spotify"`  (this should skip the metafile creation)  then have a method in python to loop and check (if exist) and send the signal to the PID inside `${XDG_RUNTIME_DIR}/hyde/mediaplayer_{player_name}.pid`.


4. Lastly the cleanup.  Be sure to only create 1 pidfile for 1 player. And this can be deleted  

 let me know if it is feasible and you are comfortable doing this.

heeeeeeeeeeh:
This is a cool idea. some questions though:
1.  For the PID file would we only get the first line and signal that process? 
2. if the user specifies multiple players  should the file be like `mediaplayer_{player1name}-{player2.pid}`? 
3. If the user didn't specify at all I think the PID should  be written to  `mediaplayer.pid`

kRHYME7:
1.  Let's just loop it. As we know waybar invokes multiple mediaplayer.py instances. Therefore we can use the player name as a key and the contents will be a list of PID, we can loop to send signals.
We can also add checks if pid still exist, if not we can pop it out in the pid file instead of sending a signal.

Or gather all existing PID and signal to those, the stored active PID list will override the PID file.


2. If the user specifies multiple players then it creates another PID file. mediaplayer_{player2}.pid

Note that a PID file can store multiple PIDs


3. If user does not specify either name it `mediaplayer.pid` or `mediaplayer_default.pid` to conserve the {player} convention.

kRHYME7 (approved):
will refactor this soon using sockets

heeeeeeeeeeh:
Thanks couldn't find the time to continue working on this rn

kRHYME7:
Hi @heeeeeeeeeeh  I really forgot that we use playerctl therefore this mediplayer can be controlled using playerctl! Should be fine right now. Thank you!

heeeeeeeeeeh:
Cool

kejioz:
my mediaplayer.py is missing from $HOME/.config/waybar so the custom/media waybar module is not working properly. Am i supposed to manually get one?
image
image

heeeeeeeeeeh:
the custom/mediaplayer module should be executing [hyde-shell mediaplayer.py](https://github.com/HyDE-Project/HyDE/blob/master/Configs/.local/share/waybar/modules/custom-mediaplayer.jsonc). Maybe update. Install hyde-cli-git and run:
- `Hyde restore Link <Path to local git repo>`
- `Hyde update`
- `Hyde upgrade`

After updating try to run `hyde-shell mediaplayer.py` and if it gives you error could not find module gi run `hyde-shell pyinit`

PR #898, opened 2025-06-28
fix: do not use uwsm for hyprlock

zeerayne:
# Pull Request

## Description

This PR fixes flaky bug when sometimes if hyprlock was launched using uwsm, after unlocking hyprlock is started again forcing user to unlock one more time. 

Maybe bug is related to `Restart=on-failure` option in unit file used by uwsm and non-zero exit code by hyprlock, but I can not reproduce the bug stable enough.

Solution is not to use uwsm for launching hyprlock, if uwsm is not used - bug is not present in any scenarios.

## Type of change

## Checklist

### DIFF
--- a/Configs/.local/lib/hyde/hyprlock.sh
+++ b/Configs/.local/lib/hyde/hyprlock.sh
@@ -320,7 +320,7 @@ if [ -z "${*}" ]; then
     print_log -sec "hyprlock" -stat "setting" " $HYDE_CACHE_HOME/wallpapers/hyprlock.png"
     "${scrDir}/wallpaper.sh" -s "$(readlink "${HYDE_THEME_DIR}/wall.set")" --backend hyprlock
   fi
-  uwsm app -- hyprlock || hyprlock
+  pidof hyprlock || hyprlock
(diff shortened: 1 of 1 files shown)


kRHYME7:
maybe hypridle fires it again?

try killall hypridle then lock it.

zeerayne:
> maybe hypridle fires it again?
> 
> try killall hypridle then lock it.

Yeah, I have tried a lot for two weeks before localizing the problem:
1. Only manual lock without `hypridle` listener
2. Modifying `lockscreen.sh` because I assumed if `hyprlock.sh` returns non-zero exit code it can trigger `hyprlock`
3. Running `hyprlock.sh` directly instead of `lockscreen.sh`
4. Running `hyprlock` executable directly - at this stage problem dissapears, so I assumed that there is smth in `hyprlock.sh`
5. After some iterations I have figured out that problem reproduces only if `hyprlock` is launched via `uwsm`

p.s. this take me a lot of time because problem do not reproducing stable

kRHYME7:
Thank you!

PR #909, opened 2025-06-30
backport (shaders): Shaders v300

kRHYME7:


kRHYME7:
@mahaveergurjar  try

mahaveergurjar:
okay

mahaveergurjar:
give this error 
image

kRHYME7:
@mahaveergurjar  

`shaders.sh --select` 

output?

kRHYME7:
Look this is the output of `shaders.sh` it should reflect in your output too. 

shaders2
shaders

mahaveergurjar:
now working

mahaveergurjar:
by the way add version on the first line so not need to write in script to detect the version 

image

kRHYME7:
> by the way add version on the first line so not need to write in script to detect the version

it's fine but you can do a PR in dev. 

It is being precompiled to cater sourcing from a different file like `file.inc` for overrides because it does no handle  multiple files. 

Also the actual compiled shader is in `~/.config/hypr/shaders/.compiled.cache.glsl` this is to ensure it is easier to debug later.

mahaveergurjar:
ok

PR #932, opened 2025-07-03
DND toggle 

mahaveergurjar:
# Pull Request

## Description
add dnd toggle keybind

## Screenshots

image


amit-0i:
i think this is already available in `custom/notifications` module in waybar

edit:- it's renamed to `custom/dunst` now?

PR #978, opened 2025-07-13
optimize picker + fuzzy search

prime-run:
This PR optimizes emoji and glyph pickers by:

- shifting `uniqueness` filter to `rofi`
- streams from `*.db` files to the picker
- and other minor refactors

Also this PR fixes bugs regarding `glyph` db file where for some glyphs extra whitespace `U+0020` would change the size and or other missbehavior in some parsers (more context below). 

And also updates and fixes `glyph.db` changing the spacer to `\t` and also removing the `nf-*` prefix from enteries since there are `fa-*` variants included.

And also, improves both pickers by enabling `fuzzy` search for both pickers, (e.g. previously searching for `cheri` would filter out `chevron_right` glyphs while it's a perfect candidate if `fuzz` is applied.

---------------

Contex: previous `glyph.db` (`blame: 76b05005 kRHYME7 (2024-12-16 00:46):
feat: animation selector won't reload entire compositor`) had `1772` enteries more than current one which is generated from [nf glyphnames.json](https://github.com/ryanoasis/nerd-fonts/blob/master/glyphnames.json) using this [custom script](https://gist.github.com/prime-run/a844c2b82f4c2c883402a88b6d4f9698), now while `diff` tool breaks if we don't filter out spaces, we can sort new .db file agains the old one and dump extras. yet, extra's and diff of the soreted db agains old one includes ~500 enteries! 
we could count them for updates in that nf source, but almost every entery in that 500 list, could be grepped in both files! (filterd for similarities)

So the only thing that is left is at least in the that list of `500` we had `Unicode` curroption, causing issues like the screenshot below (there is no whitespace typed in the edit buffer or the configuration string, it's the glyph adding 4 extra spaces)

image 

--------------------

P.S : it was a nightmare to debug XD

image

--------------

@kRHYME7 you had some TODO comments in `glyph picker script` regarding updating the `.db` file, I suggest avoiding it on user side because what ever method we come up with depends on `nf` specifications and an update on their side might cause our method to fail hence picker fails! 

I suggest: either do it manually every now and then and just update the file as this PR, or we can have an small `workflow` that updates it on each `nf` release (or updates it before `Hyde` release)


kRHYME7:
Read y to merge?

prime-run:
> Read y to merge?

I think so!

prime-run:
btw, c9c5ccc , I always rebase against target before PR, you prefer merge on this repo ?

kRHYME7:
As long as it's in dev branch.

prime-run:
> As long as it's in dev branch.

I rebase agains the target! not the target if there are more `devs` than me on that branch!
(meanwhile i have rebased master several times on hyde-ipc repo 🤣🤣)

prime-run:
> you had some TODO comments in glyph picker script regarding updating the .db file,

and what do you wanna do regarding this one ?

honestly i think manually updating it is the best choice!

kRHYME7:
> > you had some TODO comments in glyph picker script regarding updating the .db file,
> 
> and what do you wanna do regarding this one ?
> 
> honestly i think manually updating it is the best choice!


Yeah, As long as someone can monitor or routinely update it then why not. Do NF hav release cycle? maybe we can make a CI for this and creates a PR every some time.

kRHYME7:
~Hey BTW! give me back my multi select~ NVM I'll add a add an extended arg option for this.

prime-run:
> Hey BTW! give me back my multi select

I actually had a `while` loop and removed it and squashed the commit last second! 


[opts for emoji*.sh and fix multi select in glyphpicker](https://github.com/HyDE-Project/HyDE/pull/978/commits/baf86b0b1366f364b76c3eb857b2a8e9070aa3f0)

PR #1004, opened 2025-07-16
feat: prompt to disable hyde-config and optional zsh plugins

marcoscannabrava:
## Description

Implements a user prompt for disabling hyde configuration auto-reloading. In my fresh Arch installation I ran into an issue where the script was broken so I chose to disable it instead of investigating the bug*.

Adds mostly optional ZSH plugins.

Minor cleanup duplicate package: `starship`.

> *error was something along the lines of `failed to connect to user scope bus via local transport` related to `DBUS_SESSION_BUS_ADDRESS` or `XDG_RUNTIME_DIR`

## Type of change

- [X] **New feature** (non-breaking change which adds functionality)

## Checklist

- [X] I have read the [CONTRIBUTING](https://github.com/HyDE-Project/HyDE/blob/master/CONTRIBUTING.md) document.
- [X] My code follows the code style of this project.
- [X] My commit message follows the [commit guidelines](https://github.com/HyDE-Project/HyDE/blob/master/COMMIT_MESSAGE_GUIDELINES.md).
- [X] I have added necessary comments/documentation to my code.
- [X] I have tested my code locally and it works as expected.
- [X] All new and existing tests passed.


kRHYME7:
> In my fresh Arch installation I ran into an issue where the script was broken so I chose to disable it instead of investigating the bug*.

You're not addressing the error. That daemon is vital. I don't think I can add it upon installation.

Users can always disable that after installation.


BTW, reopen that issue if it persists.


I want to make it clear that this hyde-config is vital but It's fine to disable it only after installation or user can remove that on the svc.lst which disables it upon installation. That's how to avoid service from starting.

PR #1015, opened 2025-07-18
feat: Added a minimal waybar, for fans of minimalism

hell-soon:
# Pull Request

## Type of change

## Checklist

## Additional context

I often use this lix, and I often have to reinstall it by configuring waybar and other applications. You can add my waybar. It is minimized as much as possible and without unnecessary modules.

### DIFF
--- a/Configs/.local/share/waybar/layouts/hellsoon.jsonc
+++ b/Configs/.local/share/waybar/layouts/hellsoon.jsonc
@@ -1,0 +1,64 @@
+{
+    "layer": "top",
+    "output": [
+        "*"
+    ],
+    "position": "top",
+    "mode": "dock",
+    "height": 10,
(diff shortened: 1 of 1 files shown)


kRHYME7:
Hi @hell-soon  Users can always add their  own configuration layouts in `~/.config/waybar/layouts/`

then run ` hyde-shell waybar --select` to select the layout. 

IMO, this layout is not very different of the existing `hyprdots` layouts we have. 


---


However  I will  be happy if someone can add a new style for waybar, something  that is generic enough that  could fit with most the existing layouts.  As we have tons of layouts here for users to copy and configure but we lack css styles. 


#### Our layouts 
image

#### Our style selection 

also hyprdots and defaults are just the same and `macos` style is too simple. 
image

PR #1077, opened 2025-08-05
fix: adapt `screenshot.sh` for updated grimblast script

zeerayne:
This is just a copy of https://github.com/HyDE-Project/HyDE/pull/1049

Looks like commit was dropeed in recent merges and area screenshot isn't working again


kRHYME7:
Thank you @zeerayne  !!!

PR #1100, opened 2025-08-10
fix: do not escape `;` symbol in QT_QPA_PLATFORM env var

zeerayne:
Looks like part of changes is missing after megres. 

This PR is a copy of https://github.com/HyDE-Project/HyDE/pull/1078


kRHYME7:
mybad

PR #1129, opened 2025-08-13
Fix symlinks

### LINKED ISSUE #1128: [Bug]: Some scripts don't work when using symlinked files in .config
### Description

I use stow to symlink my dotfiles and some scripts dont work as they use `find` without `-follow`, I fixed these with a PR in some scripts but I believe there are more scripts with this patetrn and during each update some new script breaks. Would be nice to have some sort of universal fix, or contribution guideline about using `-follow` when adding and `find`

Ref: https://github.com/HyDE-Project/HyDE/pull/924
Ref: https://github.com/HyDE-Project/HyDE/pull/887

### Steps to Reproduce

1. Use symlinked .config, for example shaders config
2. Reload with hydectl
3. You will start seeing hyprland error messages

### Expected Behavior

Should work with symlinks

### Actual Behavior

Broken

### Screenshots

_No response_

### Environment

Hyprland 0.50.1 built from branch  at commit 4e242d086e20b32951fdc0ebcbfb4d41b5be8dcc  ([gha] Nix: update inputs).
Date: Sat Jul 19 21:37:06 2025
Tag: v0.50.1, commits: 6291
built against:
 aquamarine 0.9.2
 hyprlang 0.6.3
 hyprutils 0.8.1
 hyprcursor 0.1.12
 hyprgraphics 0.1.5

### Additional Information

_No response_

deepu105:
Fix #1128

### DIFF
--- a/Configs/.config/hyde/wallbash/scripts/code.sh
+++ b/Configs/.config/hyde/wallbash/scripts/code.sh
@@ -10,14 +10,14 @@ fi
 
 cacheDir="${cacheDir:-$XDG_CACHE_HOME/hyde}"
 # confDir="${confDir:-$XDG_CONFIG_HOME}"
-readarray -t codeConf < <(find "$confDir" -mindepth 1 -maxdepth 1 -type d -name "Code*" -o -name "VSCodium*" -o -name "Cursor*" | sort)
-readarray -t codeVsix < <(find "$HOME" -mindepth 1 -maxdepth 1 -type d -name ".vscode*" -o -name ".cursor" | sort)
+readarray -t codeConf < <(find -L "$confDir" -mindepth 1 -maxdepth 1 -type d -name "Code*" -o -name "VSCodium*" -o -name "Cursor*" | sort)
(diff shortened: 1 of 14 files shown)


kRHYME7:
Please push this to dev

kRHYME7:
also We need other users to test this. In the hyprdots era we use symlink. Then in early state of HyDE we avoided that.

But I am open to try this and seeing how we should adapt.

@HyDE-Project/testers @richen604 @rubiin @zeerayne @dieBakterie  anyone?

zeerayne:
@kRHYME7 Will test it today or tomorrow

kRHYME7:
tyt

richen604:
i already have [this change](https://github.com/richen604/hydenix/blob/3092957339803f310c13c85d485eeffa84cf5ec7/hydenix/sources/hyde.nix#L25) polyfilled in hydenix land so this looks good to me

kRHYME7:
Yeah I think we can just fix what's broken later then lolz. I'll accept this

kRHYME7:
Regression in theming and caching. I will use `-H` in theming part, this means the input command line will be resolved but anything that traverse will be treated as a file.

PR #1134, opened 2025-08-14
Fix: Uwsm changes

Olaipai:
# Pull Request

## Description
My first PR in a decade so excuse me if I am missing something.
Changes to UWSM to adhere to documentation.
Moved compositor variables to env-hyprland.d and all others to env.d. 

[Hyprland Wiki](https://wiki.hypr.land/Configuring/Environment-variables/)
```
…Use ~/.config/uwsm/env for theming, xcursor, Nvidia and toolkit variables, and ~/.config/uwsm/env-hyprland for HYPR* and AQ_* variables.
```
[UWSM Doc](https://github.com/Vladimir-csp/uwsm?tab=readme-ov-file#4-environments-and-shell-profile)
```
For uwsm-managed graphical session: export in ${XDG_CONFIG_HOME}/uwsm/env
For uwsm-managed graphical session of specific compositor: export in ${XDG_CONFIG_HOME}/uwsm/env-${desktop}
```

[Freedesktop Entry XDG Base Dirs](https://specifications.freedesktop.org/basedir-spec/latest/)

Removed XDG base directory envs as these are already set by UWSM following the XDG data hierarchy. 

Cleaned up files, easier to track and for consistency and possible issues. 

## Type of change

- [x] **Technical debt** (a code change that does not fix a bug or add a feature but makes something clearer for devs)

## Checklist

## Additional context
After pulling and installing changes (been a while since I last had done so) I noticed so odd stuttering or odd frame times. I noticed the new UWSM folder and files. Made proposed changes here and everything was smooth sailing once again. This was probably more so that the repo files clashed with my local ones. I didn't dig too deep here.


kRHYME7:
We wanted to make


```
> # Toolkit Backend Variables - https://wiki.hyprland.org/Configuring/Environment-variables/#toolkit-backend-variables 
 # GDK_BACKEND="${GDK_BACKEND:-wayland,x11,*}"       # GTK: Use wayland if available. If not: try x11, then any other GDK backend. 
 # SDL_VIDEODRIVER="${SDL_VIDEODRIVER:-wayland}" # Run SDL2 applications on Wayland. Remove or set to x11 if games that provide older versions of SDL cause compatibility issues 
 # CLUTTER_BACKEND="${CLUTTER_BACKEND:-wayland}" # Clutter package already has wayland enabled, this variable will force Clutter applications to try and use the Wayland backend 
  
 # # Qt Variables  - https://wiki.hyprland.org/Configuring/Environment-variables/#qt-variables 
  
 QT_QPA_PLATFORM="${QT_QPA_PLATFORM:-wayland;xcb}" # Qt: Use wayland if available, fall back to x11 if not. 
 QT_AUTO_SCREEN_SCALE_FACTOR="${QT_AUTO_SCREEN_SCALE_FACTOR:-1}"                 # Enables automatic scaling, based on the monitor’s pixel density 
 QT_WAYLAND_DISABLE_WINDOWDECORATION="${QT_WAYLAND_DISABLE_WINDOWDECORATION:-1}" # Disables window decorations on Qt applications 
 QT_QPA_PLATFORMTHEME="${QT_QPA_PLATFORMTHEME:-qt6ct}"                           # Tells Qt based applications to pick your theme from qt5ct, use with Kvantum. 
  
 # # HyDE Environment Variables - 
  
 MOZ_ENABLE_WAYLAND="${MOZ_ENABLE_WAYLAND:-1}"                        # Enable Wayland for Firefox 
 GDK_SCALE="${GDK_SCALE:-1}"                                          # Set GDK scale to 1, for Xwayland on HiDPI displays 
 ELECTRON_OZONE_PLATFORM_HINT="${ELECTRON_OZONE_PLATFORM_HINT:-auto}" # Set Electron Ozone Platform Hint to auto, for Electron apps on Wayland

```

these as hyprland specific. Assuming that we might run other dekstop environment with uwsm.

kRHYME7:
The reason I did not add the toolkit env inside env.d is because it populates it no matter what session you're using.


So for example I want all the QT_* vars exclusively for Hyprland session I can do that. Selecting a KDE session will not conflict. That's the reason I moved all the toolkit specifics to Hyprland's 


I tried this with niri, Hyprland and kde. Setting envs explicitly per session feels reassuring.


---

Another example is because niri doesn't have an xwayland natively I used xwayland-satellite and I want to have different toolkit settings across sessions.

kRHYME7:
About the xdg stuff removal, yeah that's fine.

I'm aware of UWSM setting it for you but added it for redundancy.  But it might cause confusion so I'm pleased you removed that. 👍👍👍

kRHYME7:
>  I noticed so odd stuttering or odd frame times. 

is this caused by envs being added in env-hyprland?

Olaipai:
> The reason I did not add the toolkit env inside env.d is because it populates it no matter what session you're using.
> 
> So for example I want all the QT_* vars exclusively for Hyprland session I can do that. Selecting a KDE session will not conflict. That's the reason I moved all the toolkit specifics to Hyprland's
> 
> I tried this with niri, Hyprland and kde. Setting envs explicitly per session feels reassuring.
> 
> Another example is because niri doesn't have an xwayland natively I used xwayland-satellite and I want to have different toolkit settings across sessions.

I'm sorry, but I do not understand how KDE and Niri is relevant here? Is HyDE branching out? 

However, do you use UWSM with KDE and Niri? I know nothing of Nirir and haven't touched KDE for years now. 
If you don't use UWSM and and want to make sure you get a clean slate / clean dbus on log out, you need to use use ``` loginctl terminate-user ""```. That should clean up the exit and make way for a new login session with diffs vars and not have residual vars from UWSM / D-bus. 
If you use `uwsm stop` you might not get a clean end and D-bus seems to be unaffected and retains information such as env vars.
If you end with `hyprctl dispatch exit 0` you're hijacking the session out from underneath UWSM and is messy as all hell lol. Just don't!

If you're using UWSM with only Hyprland (which I would assume 'cuz Hyprland DE), then your argument is moot since no other DE / WM is using UWSM and ending your session properly should not affect any possible logins from a greeter / tui, regardless of which file said toolkit is declared. 

Putting the toolkit vars etc in the `env-hyprland.d` section is akin to putting them into the `hyprland.conf` which is explicitly stated you should avoid doing. Why? I would assume because these has nothing to do with the actual compositor and we are now actively handing of session management to UWSM. And since our main styling is done in the `*.conf` files for said compositor we're letting Dbus handle the toolkit side of things. __This is my GUESS! I have no in-depth knowledge of this lol__ 

If I misunderstood your point here, please correct me. :smile: 

The primary reason for splitting them up into the correct folder are stated in the docs – Let the compositor do compositor things and let the session manager deal with session things like D-bus, systemd etc. 
 ___

> > I noticed so odd stuttering or odd frame times.
> 
> is this caused by envs being added in env-hyprland?

I believe this was some sort of clash between what I had set (which was in the env) clashed with what was in env-compositor. I do not remember exactly which variable I found caused this I believe it was something related to the electron hint and the firefox wayland trigger clashing with something I had set locally on my system;  I just remember seeing errors from UWSM in journal. Only thing I changed was the HyDE files as proposed here and my issues went away. I am not 100% sure if this was the cause or just some odd correlation which happened to fix it at the same time. This was maldening since changing focus with the mouse meant half a second of stuttering :anger: 
I did however confirm some off scaling issues, since I had GDK scaling set to 2, so one set was declaring 2 the other declaring 1. I am not sure if this is actually possible, but some apps were scaled accordingly, some were so tiny I had to whip out a magnifying glass :rofl: 
___

> About the xdg stuff removal, yeah that's fine.
> 
> I'm aware of UWSM setting it for you but added it for redundancy. But it might cause confusion so I'm pleased you removed that. 👍👍👍

Yes, this confused me for a second and was exactly why I removed it lol :rofl:

Olaipai:
This is how Ihave my logout set in `wlogout` layouts. Everything gets shutdown cleanly and checking in tty after logging out shows no residual values set in `env.d` from the Hyprland session.
If I use `uwsm stop` I can see residual env vars from `env.d` in a tty session. Which if I understand correctly is NOT what you want since you have different toolkit vars when using Niri or KDE, right? 

``` 
{
    "label": "logout",
    "action": "loginctl terminate-user ''",
    "text": "Logout",
    "keybind": "e"
}
```

Olaipai:
@kRHYME7 Is there any reason why `APP2UNIT_SLICES` variable is present in "uwsm/env"? The necessary package for this variable is not present in any `.lst` anywhere AFAICT. Should remove it or add `app2unit-git` as a package. [app2unit](https://github.com/Vladimir-csp/app2unit)
```
export APP2UNIT_SLICES='a=app-graphical.slice b=background-graphical.slice s=session-graphical.slice'
```

kRHYME7:
> @kRHYME7 Is there any reason why `APP2UNIT_SLICES` variable is present in "uwsm/env"? The necessary package for this variable is not present in any `.lst` anywhere AFAICT. Should remove it or add `app2unit-git` as a package. [app2unit](https://github.com/Vladimir-csp/app2unit)
> ```
> export APP2UNIT_SLICES='a=app-graphical.slice b=background-graphical.slice s=session-graphical.slice'
> ``` 


I bundle it together with HyDE

kRHYME7:
> If I misunderstood your point here, please correct me


https://github.com/Vladimir-csp/uwsm?tab=readme-ov-file#4-environments-and-shell-profile

I think you missed this part. `env` and `env-compositor` are like `.bash_profile`/`.zprofile` but loaded after the graphis. Check bullet 3 and 4.  I interpreted this as shell agnostic exporting of variables.


Reasons why those variables are located there:


1. UWSM - is a Universal Wayland Session Manager -- not Hyprland only

2. One can use uwsm select to select the session. This will/can use `uwsm start -- default` that means if you add variables  in `env` it will populate to every session. Note some reportedly ditch sddm as they use uwsm. 

3. The point of managing the session is to isolate some environments vars. There's no magic going on, it is only exporting env variables at the right time. Some of the toolkit env conflicts with other desktop sessions like kde. In kde you do not want to set QT_* or else some apps becomes problematic. Other toolkit envs though can go to env, but I realized I'll just bundle it together in `env-hyprland` together with QT_*

4. We are not branching  out but I don't want HyDE to mangle with other sessions unnecessarily. So I actually run HyDE with niri and separating env-hyprland and env-niri vars explicitly.  Even some are redundant variables but we cannot be sure which vars are not compatible, also helps me test. Choosing kde plasma session in uwsm select will not conflict env vars.

Olaipai:
Excuse me if I am completely misunderstanding here, been under the weather for a week now :rofl: 

>UWSM - is a Universal Wayland Session Manager -- not Hyprland only. 
Some of the toolkit env conflicts with other desktop sessions like kde. In kde you do not want to set QT_* or else some apps becomes problematic

– Don't use UWSM with KDE or any DE. It's redundant. DEs inherently manages itself. UWSM wraps the compositor into units/slices to make it act more like a DE instead of this one monolithic unit.

---

> I think you missed this part. env and env-compositor are like .bash_profile/.zprofile but loaded after the graphis. Check bullet 3 and 4. I interpreted this as shell agnostic exporting of variables.


– Normally I would agree with you here, but my interpretation is different and here is why, the Hyprland wiki states explicitly:

`Use ~/.config/uwsm/env for theming, xcursor, Nvidia and toolkit variables, and ~/.config/uwsm/env-hyprland for HYPR* and AQ_* variables. `
 
And on the UWSM repo it states on bullet point 3 & 4:

>For __<ins> uwsm-managed graphical session: </ins>__ export in ${XDG_CONFIG_HOME}/uwsm/env

>For uwsm-managed graphical session of __<ins> specific compositor </ins>__: export in ${XDG_CONFIG_HOME}/uwsm/env-${desktop}

These points is exactly what led me to make the changes I did and this PR.

Edit: After some thinking, the already implemented solution you have also does make sense as well. And there might not a be a "correct" implementation here.

---

I know nothing of Niri, but to me for the scope of UWSM this seems like a good thing for consistency across compositors. Until some compositor might not play nice I guess. 

>So I actually run HyDE with niri and separating env-hyprland and env-niri vars explicitly. Even some are redundant variables

– I don't know much if anything about Niri, just had a quick look but I mean… You do kinda reinforce the point: Use `env` for theming, toolkit, nvidia and `env-compositor` for compositor specific vars. I guess your solution would work fine here, I don't know enough about the backend of either UWSM or Hyprland etc to know what the pros and cons of this implementation. 
My intrinsic feeling is that one is just a general way of managing the session adhering to the general way of managing a graphical Wayland session, while the latter is more of tailored way to work with a specific compositor (Sway, Wayfire, Hyprland, Labwc)
And this seems to make sense to me as we want the session manager to handle the vars and not the compositor, hence why they are separated. 
Taking a look at  [Section 2](https://github.com/Vladimir-csp/uwsm?tab=readme-ov-file#2-service-startup-notification-and-vars-set-by-compositor) on the UWSM repo, this reinforce my thoughts (I think, I might be too dumb :rofl:)

If you're talking about running multiple sessions of different compositor / DE instances, you're almost bound to end up with conflicts tho and you're better of with a T2 hypervisor or w/e VM/VE of your choosing

kRHYME7:
> If you're talking about running multiple sessions of different compositor / DE instances, you're almost bound to end up with conflicts tho and you're better of with a T2 hypervisor or w/e VM/VE of your choosing


I'm talking about switching different DE/Sessions where I want to specifically declare what Envs goes in there.

kRHYME7:
I'll dodge the points referencing Hyprland wiki about uwsm vars. Because I think it is generalizing for simplicity. It is Truthful that HYPR and AQ vars should be in env-hyprland and others can be in env. 

I want to use UWSM flexibly and I don't think it will give errors. As UWSM promises to separate per session envs. Being said, I think it is better to add per toolkit stuff inside env-hyprland too, for now, as we can't be sure how other sessions behave with those toolkit. Especially those sdl stuff.


As said some uses `uwsm select` to enter the session without relying on SDDM.


The current implementation also Isolates it pretty well.

PR #1291, opened 2025-09-18
Update cat alias to use default bat behavior

### LINKED ISSUE #1290: Modify cat alias to use bat's default settings
**Current Behavior**  
If `bat` is installed, we currently set the following alias:
```bash
alias cat='bat --style=plain --paging=never --color auto'
```
This forces plain output without paging, and explicitly sets --color auto (which is already the default).

**Problem**
Some users, including myself, prefer using bat with full styling and a pager. This alias overrides those preferences, making it inconvenient to work with the default bat features.

**Proposed Change**
Update the alias to:
```bash
alias cat='bat'
```
This way, users will get the default bat behavior, including their preferred styling and paging settings.

std-enigma:
Closes #1290


kRHYME7:
It breaks cat command on scripts sometimes. 

The reason why zsh is using that too 

image
 
Feel free to override this behaviour outside HyDE configs.

std-enigma:
Can't we just explicitly add the required arguments in the scripts? instead of the alias itself

PR #1316, opened 2025-09-30
feat: add pipewire support for `volumecontrol.sh`

zeerayne:
# Pull Request

## Description

This feature allow to manage volume using native pipewire utilities, if audioserver is running by PipeWire.

If volume is managed throug Pulse Audio when the actual audio server is pipewire this leads to audio stream is recreated every time any parameter is changed:
- muted
- unmuted
- volume changes

This behaviour leads to some inconvenience:
- Any equalizer/filter/etc applied to physical sink is discarded when stream is recreated
- Waybar crashes on eq discard for some reason
- Virtual sink should be used as workaround

Managing volume natively by pipewire utils solves all mentioned problems.

## Type of change

## Checklist

## Additional context

> [!WARNING]
> I have tested it only on my PC and my laptop, which are both AMD-only and have pretty similar audio configs. I will appreciate if someone can test this on different configurations

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 
 ### Fixed
 - Hyprland: Fix errors when `HYPRLAND_CONFIG` is not set yet
-- Fish: Pleas Move you configs to `~/.config/fish/conf.d` 
+- Fish: Please Move you configs to `~/.config/fish/conf.d`
 
 
 ### Added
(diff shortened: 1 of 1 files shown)


kRHYME7:
is this a feat or a sort of fix? Assuming  HyDE uses full pipewire right. 

So this means this is the proper way to manage pipewire volume?

kRHYME7:
@zeerayne Looks Good! please add CHANGELOG entry please

kRHYME7 (approved):

zeerayne:
> is this a feat or a sort of fix? Assuming HyDE uses full pipewire right.
> 
> So this means this is the proper way to manage pipewire volume?

Yeah, some kind of "more proper way to manage volume if pipewire server is running". 
* By default `volumecontrol.sh` manage volume with Pulse utils, which leads to pipewire streams to be recreated when volume changes.
* But when stream is recreated it discards all other settings from other sources (i.e. Equalizer)
* Why it is recreated - looks like some internal logic in `pipewire-pulse` which can not be controlled externally
* If volume is managed via pipwire-native utils, streams are not recteated, but amended instead, which is looks like more proper way and did not show any sideeffects during testings

zeerayne:
> @zeerayne Looks Good! please add CHANGELOG entry please

Sure, will do it

kRHYME7:
Thanks @zeerayne

PR #1370, opened 2025-10-19
(It is HyDE btw) Hyde cursor when in hyprlock lockscreen

Saphereye:
# Pull Request

## Description

Added the following snippet to the Hyprlock configuration to hide the cursor on the lock screen:
```css
general {
    hide_cursor = true
}
```

The cursor was previously visible during lockscreen, which felt a little unintuitive. This change hides it.

## Type of change

- [X] **New feature** (non-breaking change which adds functionality)

## Checklist

- [X] I have read the [CONTRIBUTING](https://github.com/HyDE-Project/HyDE/blob/master/CONTRIBUTING.md) document.
- [X] My code follows the code style of this project.
- [X] My commit message follows the [commit guidelines](https://github.com/HyDE-Project/HyDE/blob/master/COMMIT_MESSAGE_GUIDELINES.md).
- [X] I have tested my code locally and it works as expected.
- [X] All new and existing tests passed.

## Screenshots

I am trying to capture the cursor in satty but I am a little stuck there. I will update the PR if it works out.

## Additional context

An issue was reported in hyprlock where the cursor wouldn't be visible after logging back in: https://github.com/hyprwm/hyprlock/issues/494. However, I was unable to reproduce that behavior locally, and these changes appear to work as intended.


rubiin (approved):

kRHYME7:
Oh no it breaks some lockscreen with media controls

Saphereye:
@kRHYME7, I think we have the following options:
1. Revert my `hide_cursor` changes.
2. Apply `hide_cursor` selectively to configurations without media controls.
3. Use [inactive_timeout](https://wiki.hypr.land/Configuring/Variables/#cursor) as a global setting to hide the cursor after a period of inactivity. (https://github.com/hyprwm/hyprlock/issues/67)

I feel option 2 is the least invasive, but it doesn’t fully resolve the issue. Option 3 does, and I might use it personally, though I’d like to hear your thoughts on how it might feel from a user’s point of view.

I feel the best scenario would be cursor hidden by default in the lock screen, and temporarily visible on mouse movement. But I haven't found any leads on how I can do that.

kRHYME7:
I usually let my cursor hide after some time of inactivity. 

We can do 2 but then I realized it will be more intuitive to control cursor visibility in compositor level. I think we should revert this change for now


Users can just add their own `cursor:inactive_timeout ` which is way predictable.

> I feel the best scenario would be cursor hidden by default in the lock screen, and temporarily visible on mouse movement. But I haven't found any leads on how I can do that.

If only there's a way to hide the cursor without any change in state, that would be helpful.

Saphereye:
@kRHYME7 That makes sense. I think we should go ahead with reverting it :+1:

kRHYME7:
I'll probably add a workaround during `hyprlock.sh` calls. Let's see what we can do.

Saphereye:
> If only there's a way to hide the cursor without any change in state, that would be helpful.

I think we can achieve this by editing [this](https://github.com/hyprwm/hyprlock/blob/36ec73f166d9434a3f27c96c575198906f77644a/src/core/Seat.cpp#L33) part of the code. It does have the skeleton to get mouse movements and hiding/unhiding the cursor. We could add a config option to temporarily unhide the cursor and then restore the previous behavior.

The only downside is that we’d be cracking open Hyprlock to make this work :upside_down_face:

kRHYME7:
I would rather  ask hyprland to handle the cursor stuff. But a lock screen client is a special case so maybe we can add a PR in hyprlock to handle hide-cursor upon rendering then a mouse movement re exposes  the cursor. 

edit

Or maybe we can adapt the cursor timeout in hyprlock directly.

PR #1383, opened 2025-10-25
Revert "(It is HyDE btw) Hyde cursor when in hyprlock lockscreen"

kRHYME7:
Reverts HyDE-Project/HyDE#1370


kRHYME7:
As discussed

PR #1385, opened 2025-10-25
[UWSM] Start processes as service instead of scope by default

JunaidQrysh:
# Pull Request

## Description
Change default process startup to service as it is recommended in the [official systemd docs](https://systemd.io/DESKTOP_ENVIRONMENTS/#:~:text=Using%20.service%20units%20instead%20of%20.scope%20units%2C%20i.e.%20allowing%20systemd%20to%20start%20the%20process%20on%20behalf%20of%20the%20caller%2C%20instead%20of%20the%20caller%20starting%20the%20process%20and%20letting%20systemd%20know%20about%20it%2C%20is%20encouraged.) and also followed by kde plasma(see screenshot for proof).

## Type of change

## Checklist

## Screenshots
kde_screenshot

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ All notable changes to `HyDE` will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/), and this project adheres to _Modified_ [CalVer](https://calver.org/). See [Versioning](https://github.com/HyDE-Project/HyDE/blob/master/RELEASE_POLICY.md#versioning-yymq) For more info
 
-## v25.11.1
+## Upcoming Release
 
 ### Fixed
 
@@ -13,6 +13,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 
 ### Added
 
+- Rofi: Added search functionality to the game launcher menu.
 - Cliphist: image-history #1360
 - Cliphist: Rofi binds #1360
 - Gamelauncher: lutris inspector py script now uses the lutris DB to get meta dat making it faster than using lutris CLI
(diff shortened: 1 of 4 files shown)


kRHYME7:
Thanks!

But I want to note that all waybar modules are using scope. this way if waybar dies all child process are cleaned.

Even though I explicitly use scope, maybe I forgot something so can you also check the modules please? 🥺🙏

JunaidQrysh:
I tested it , no problems as expected since the modules are manage by waybar itself and systemd waits for waybar to cleanly terminate(point of a service, that's why it is recommended)

kRHYME7:
hmmm before merging this I need to test first. 


it breaks  one module in hyde (breaks more of my local modules)

an example for this is 

image


as you see, we are now required to use `-t scope` if we intend to capture the stdout like a waybar module.  Gladly it did break single module. 


in lockscreen script (invokes a lockscreen)  https://github.com/HyDE-Project/HyDE/blob/401ccca0a4d581d52f618e296d298b23f304a822/Configs/.local/lib/hyde/lockscreen.sh#L19 this breaks as even if we use `*.scope ` in the unit name `-t scope` is required. as it contradicts the `APP2UNIT_TYPE="service"` This should be updated to `.service` 



--- 

I am not saying that  this PR is bad but we need to adjust some of our script first. Can you help me find some stuff that breaks. Right now I only seen those 2 mentioned.

---

I needed help in deciding that should all the apps executed by the launcher use service? 

image

JunaidQrysh:
@kRHYME7 After more testing, it also breaks hyprlock which is a one line fix, and since app2unit is used at multiple places there could be more things to break. I think the best way is to just launch apps through rofi as service.
The env should have been there since the uwsm merge, but anyway.

kRHYME7:
I won't be releasing until 25.11.1 so We have extra time to fix those.

As of the moment I only stumbled upon those 2 regressions.


> The env should have been there since the uwsm merge, but anyway.


Yeah I should have known. I just use the defaults as provided by uwsm.

JunaidQrysh:
@kRHYME7 i think its ready :)

kRHYME7:
@rubiin @zeerayne  anyone?

kRHYME7:
Changelog please

zeerayne:
lgtm, but will test it later today

kRHYME7:
will merge this to dev now for testing

kRHYME7:
Thank you!

Issue #1424, opened 2025-11-21
[Bug]: Chase Hyprland


kRHYME7:
### Description

Hi @rubiin  @HyDE-Project/testers  Anyone of you here is using hyprland-git? We got a masive breaking change for hyprland. 

https://github.com/hyprwm/Hyprland/pull/12269

Anyone willing to open a PR, please 

### Steps to Reproduce

Use hyprland-git in AUR

### Expected Behavior

na

### Actual Behavior

na

### Environment

NA

rubiin:
looks like windowrule snytax changes

rubiin:
https://wiki.hypr.land/Configuring/Window-Rules/ docs added

kRHYME7:
Ahhhh took me an hour or two to fix all 15k syntax errors lol. Please review my PR so we can merge it to dev then Will be adding a release soon. Maybe when 0.53 comes out

PR #1425, opened 2025-11-21
Fix: Chase Hyprland 0.53

kRHYME7:
#1424 

So Hyprland will break user configs soon. Please follow the changes in here.  

Edit: 
as mentioned in https://github.com/hyprwm/Hyprland/issues/12274 we will try our best to smooth out any config issues in our part.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 
 ### Fixed
 
-- Gamelauncher: steamdeck holograph
+- Game launcher: steamdeck holograph
 - Formatting using 
 - Hyprland 0.53 Syntax Revamp
 - Some old hyprdots theme parsing. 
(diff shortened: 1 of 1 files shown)


rubiin:
should we adopt for this syntax with braces and name

```javascript
windowrule {
  name = apply-something
  match:class = my-window

  border_size = 10
}
````

**Comment by kRHYME7** [comment]
@rubiin  We only add name for windows rule that we want to dynamically turn off and on. Which is I think way way better for rules with categories. But do we really need this? ( One usecase is to turn off hyde's windows rules dynamically? Any other use case?

I am treating this as global as of now so names are not important.  Have you figured any usecase? or maybe categorize the windowrules ? Let me know

**Comment by vaxerski** [comment]
use version guards and save yourself some hassle.

https://wiki.hypr.land/Configuring/Uncommon-tips--tricks/#config-versioning

**Comment by kRHYME7** [comment]
we love you vaxx 😌

@vaxerski Will this  skip parsing/loading the configs when version mismatched?

**Comment by vaxerski** [comment]
just a conditional like any other in hyprlang, but yes this is a way to version-gate configs

**Comment by kRHYME7** [comment]
@vaxerski   I am currently testing it. Currently it is not working on hyprland-git? 
For testing I added this block 

image



This is my full test block and I am using hyprland-git 
`Hyprland 0.52.0 built from branch main at commit b5a2ef77b7876798d33502f8de006f9c478c12db clean (i18n: add Português (Portugal) translation ( 12328)).`
```


# checking if var is set by hyprland 
exec = notify-send "Test: $HYPRLAND_V_0_53" 


# hyprlang if HYPRLAND_V_0_53


exec = notify-send "confirmed: $HYPRLAND_V_0_53" 


# hyprlang endif



# hyprlang if !HYPRLAND_V_0_53


exec = notify-send "Still the old: config $HYPRLAND_V_0_53"

# hyprlang endif


```

vaxerski:
oopes, fixed in https://github.com/hyprwm/hyprlang/commit/a318deec0c12409ec39c68d2be8096b636dc2a5c

vaxerski:
0.53.0 out jsyk

kRHYME7:
Merry Christmas @vaxerski !!!

PR #1519, opened 2026-01-08
Fix:  gamemode.sh windowrule syntax issue when changing window opacity

post2424:
# Pull Request

## Description

Fixed a syntax issue in gamemode.sh that caused the script to not change the opacity of windows.

## Type of change

## Checklist

## Summary by CodeRabbit

* **Improvements**
  * Updated gamemode toggle mechanism to use persistent state management for more reliable switching between gaming and normal modes.

### DIFF
--- a/Configs/.local/lib/hyde/gamemode.sh
+++ b/Configs/.local/lib/hyde/gamemode.sh
@@ -1,7 +1,13 @@
 #!/usr/bin/env bash
-HYPRGAMEMODE=$(hyprctl getoption animations:enabled | sed -n '1p' | awk '{print $2}')
-if [ "$HYPRGAMEMODE" = 1 ]; then
-    hyde-shell workflows --set gaming
+LOCK_FILE="${XDG_RUNTIME_DIR}/hyde/gamemode.lck"
+
+if [ -f "$LOCK_FILE" ]; then
+    # Gamemode is ON → turn it OFF
+    hyprctl reload config-only -q
+    rm -f "$LOCK_FILE"
 else
-    hyde-shell workflows --set default
+    # Gamemode is OFF → turn it ON
+    mkdir -p "${XDG_RUNTIME_DIR}/hyde"
+    hyprctl keyword source "${XDG_CONFIG_HOME}/hypr/workflows/gaming.conf"
+    touch "$LOCK_FILE"
 fi


PureFallen:
I stumbled across the exact same change in `gamemode.sh` to match the new windowrule syntax.
While it does work for windows that are newly created after the script has been executed, it unfortunately doesn't seem to affect existing windows. They will still use transparency when unfocused, which I previously did not have when using HyDE. I don't know whenever that is intended behavior or not. If not, we might have to do additional changes to restore previous behavior.

kRHYME7:
I suggest using `hyde-shell workflows --select` This changes the config file explicitly.

post2424:
> I suggest using `hyde-shell workflows --select` This changes the config file explicitly.

Great suggestion.
I updated gamemode.sh to use hyde-shell workflows instead. 
Now, at least on my system, switching between workflows doesn't cause the issue described by @PureFallen.

kRHYME7:
> I updated gamemode.sh to use hyde-shell workflows instead.

gamemode.sh is  just a reference from the hyprland wiki anyway.  That's why I would still recommend our own workflows  script. This way the state is saved.

gamemode.sh will still stay for reference though. I'll try to fix some conflicts on this PR first

Configs/.local/lib/hyde/gamemode.sh
```diff
@@ -1,25 +1,7 @@
 #!/usr/bin/env bash
```
kRHYME7, line 1: Please use hyprctl for this. Sorry for the misunderstanding. Thanks!

PureFallen:
> > I updated gamemode.sh to use hyde-shell workflows instead.
> 
> gamemode.sh is just a reference from the hyprland wiki anyway. That's why I would still recommend our own workflows script. This way the state is saved.
> 
> gamemode.sh will still stay for reference though. I'll try to fix some conflicts on this PR first

Thanks for letting us know. I am still new to HyDE and try to understand some of it's inner workings.
I ended up there because `gamemode.sh` is being called by the Gamemode Hotkey (SUPER + ALT + G). So when the functionality broke for me, I ended up looking there for a solution. Little did I know about the intended solution via. workflows.

PureFallen:
Gonna bump before the PR gets closed down.
As of now this is still the most viable solution to fixing the `gamemode.sh` that is used by the hotkey.

In the meantime new people ran into the same issue and I ended up redirecting them to this PR as a workaround.
When looking further into it I came across the limitations of this method though - we are always restoring into the `default` workflow instead of whatever workflow we been in before going into `gaming`.

Depending on what the scope is we could leave it at that or look into a way to remember and restore the preview workflow. However, this could require changes in how we are detecting the `gaming` workflow since animations are also disabled for `snappy` and `powersaving`. It could perhaps be useful to look for a property that is unique to the `gaming` workflow, or to create one if necessary.

kRHYME7:
Will merge this now 
I've used a lockfile for the state and use `hyprctl keyword source gaming.conf`. 
Used transient therefore we don't need to know the current preferred workflow at the same time making gaming mode (just a gaming set of configs) be "toggle-able". 

Only the best way I can think of for now.

PR #1682, opened 2026-04-05
feat: Add optional fingerprint hyprlock installer flow

ftbhabuk:
Adds optional fingerprint unlock support for HyDE's hyprlock setup
  without changing the default layout for all users.

  This PR:
  - adds a bundled `HyDE-fingerprint.conf` hyprlock layout
  - adds an install-time prompt to enable fingerprint support
  - installs `fprintd` only when the user opts in
  - switches the initial hyprlock layout to the fingerprint layout
  - stores the selector state so the layout is remembered in the
    `SUPER+SHIFT+U` hyprlock layout menu
  - documents terminal-based enrollment and verification steps

  This avoids forcing PAM changes, since fingerprint authentication
  setup remains system- and distro-specific.

  - Implements an installer option for optional hyprlock fingerprint support. 
 ## Type of change

  ## Checklist

  - [x] I have updated the documentation accordingly.

  ## Testing

  - Verified script syntax with `bash -n Scripts/install.sh`
  - Verified script syntax with `bash -n Scripts/install_pst.sh`
  - Ran the installer in dry-run mode to confirm:
    - the new fingerprint prompt appears
    - opting in adds `fprintd` to the install flow
    - the fingerprint layout activation path is triggered

## Summary by CodeRabbit

* **New Features**
  * Optional fingerprint unlock support for hyprlock on supported devices
  * Installer now prompts to enable fingerprint authentication during setup
  * Manual configuration available for existing users

* **Documentation**
  * Added configuration guide for fingerprint unlock setup and enrollment


kRHYME7:
screenshots?  Also I htink we can put fprint for all layouts.

- in hyprland we can use $VARS and ifdef to handle fprint enable / disable. I  can cook something for that. 
- We can just add frprint non-GUI configs  in our ~/.local/share/hyde/hyprlock.conf
 
image

- Then for the GUI stuff we can just do 

```
ifdef HYDE_FPRINT_ENABLED
so something 

endif HYDE_FPRINT_ENABLE
```


But I do not think weneeded this if we only intend to use default hyprlock variables as the ywill not be shown if no fprint daemon  running

ftbhabuk:
> screenshots? Also I htink we can put fprint for all layouts.
> 
>     * in hyprland we can use $VARS and ifdef to handle fprint enable / disable. I  can cook something for that.
> 
>     * We can just add frprint non-GUI configs  in our ~/.local/share/hyde/hyprlock.conf
> 
> 
> image
> 
>     * Then for the GUI stuff we can just do
> 
> 
> ```
> ifdef HYDE_FPRINT_ENABLED
> so something 
> 
> endif HYDE_FPRINT_ENABLE
> ```
> 
> But I do not think weneeded this if we only intend to use default hyprlock variables as the ywill not be shown if no fprint daemon running

yes.BUT i thought maybe for first iteration changing all hyprlock layought maybe be a bit overkill . or we can jsut copy other lock layouts and add the additional configs for fingerprint in them too ?

kRHYME7:
it won't be overkill, those layouts will be override by HyDE anyway. Adding some lines is cheaper than adding more files (they might ask fprint support for other layouts).

What we expect is something that just works.As long as it is not a distraction for non fprint users. I don't have  fprint therefore I don't know how to test.

ftbhabuk:
> it won't be overkill, those layouts will be override by HyDE anyway. Adding some lines is cheaper than adding more files (they might ask fprint support for other layouts).
> 
> What we expect is something that just works.As long as it is not a distraction for non fprint users. I don't have fprint therefore I don't know how to test.

Then , should i open another pr with adding it in all hyprlock layouts ?

kRHYME7:
Just modify this pr and remove that newer layout

ftbhabuk:
I removed the separate fingerprint layout and moved fingerprint support into the shared hyprlock flow instead.

  I also tested it locally on my machine:
  - fingerprint unlock works with the normal HyDE layout
  - fingerprint status text works correctly
  - switching hyprlock layouts keeps fingerprint support working
  - per-layout files did not need fingerprint-specific edits

  I’ve also attached a local fingerprint enroll/verify screenshot for reference.
  - 260406_09h11m15s_screenshot

Scripts/global_fn.sh
```diff
@@ -102,6 +104,22 @@ prompt_timer() {
     echo ""
     set -e
 }
+
```
kRHYME7, line 107: this is already handled by `set_conf()`. 

https://github.com/HyDE-Project/HyDE/blob/ae5f0a70272a35cb4078171d3f7a5adb6a154cef/Configs/.local/lib/hyde/globalcontrol.sh#L255-L264


why duplicate?

kRHYME7 (changes requested):
Too much setup needed and  too much text manipulation during intallation. I have an implementation in mind. I will open a PR soon.

Configs/.local/share/hyde/hyprlock.conf
```diff
@@ -41,11 +41,30 @@ label {
...
+label {
+    text = $FPRINTPROMPT
+    color = $wallbash_txt2_rgba
+    font_size = 16
+    position = 0, 44
+    halign = center
+    valign = bottom
+}
+
+label {
```
kRHYME7, line 61: this might look weird as the fprint prompt is only in this position. 


I needed your help soon to test the correct  prompt positions.

ftbhabuk:
> Too much setup needed and too much text manipulation during intallation. I have an implementation in mind. I will open a PR soon.

okay. I'll wait. . duplicate , missed by mistake. open your pr. btw this implementation also works😊😅

Configs/.local/share/hyde/hyprlock.conf
ftbhabuk: > this might look weird as the fprint prompt is only in this position.
> 
> I needed your help soon to test the correct prompt positions.

Understood.

  I’ll pause further changes on this PR for now.

  I tested the shared fingerprint flow locally and it works, but I understand your concern about the current installation/config manipulation and the prompt positioning.

  Happy to help test prompt positions and the new implementation once your PR is up.

kRHYME7:
Hi @ftbhabuk I forgot to push it to a separate pr branch. But here is how we can add the fprint support 

https://github.com/HyDE-Project/HyDE/commit/590236f686949b623b6b9cafbac01bfddf2cd181#diff-a90baf38dba46068520a05cbad92a98fd3c0ce648eba5a47a05a7a26425aae79

So users can enable 

```toml
[hyprlock]
fingerprint_enabled = true
```

If it is enabled, an error message will be displayed in  hyprlock if command not found or fprintd does not support the fprint hardware.

At that point users can manually add fprintd or resolve any driver issues. 
 
 
 Todo: Add labels for other layouts.

PR #1684, opened 2026-04-05
enhancement: Deprecate custom pyenv and add uv as enviroment manager

### LINKED ISSUE #1683: [Feature Request]: Migrate Python environment management from custom pip+venv to uv
### The problem

HyDE scripts currently run against whatever Python environment happens to be active at call time. Some scripts activate the managed venv explicitly; others call `python3` directly and rely on `sys.path` injection at runtime to find packages. This means:

- On systems where users have their own venvs, pipx installs, conda environments, or globally installed packages, HyDE picks up whatever is in scope — behavior varies per machine.
- Package version conflicts with user-managed Python environments cause hard-to-reproduce bugs that only appear on specific setups.
- There is no guarantee that a script launched by a shell script and one launched through `hyde-shell` run against the same interpreter or package set.

### What changes with uv

uv manages HyDE's environment as a fully isolated, reproducible unit through a proper project setup:

- `pyproject.toml` declares the exact dependencies
- `uv.lock` pins every transitive dependency to a specific version
- `uv sync` converges the environment to match the lock file in a single atomic operation
- `uv add <package>` adds a new dependency and updates the lock file in one step

Every script always runs against the same locked set of packages, regardless of what else is installed on the system. Two HyDE users on different machines, or the same user after a system Python upgrade, will run identical package versions. Bugs become reproducible.

### Isolation from the rest of the system

Because every script is invoked against the managed venv's Python binary (not the ambient `python3`), HyDE is fully decoupled from:

- System-level Python packages (`/usr/lib/python3.x/`)
- User site-packages (`~/.local/lib/python3.x/`)
- Any active venv, conda env, or pyenv shim the user may have set up
- pipx-managed tools that patch `sys.path`

HyDE stops being "a collection of scripts that depend on whatever Python is in scope" and becomes a self-contained module with a defined, reproducible runtime.

### Performance

uv is written in Rust and resolves and installs packages 10–100x faster than pip. The `pyinit` step that today can take 20–30 seconds on a clean machine becomes nearly instantaneous. This matters most for first-time installs, where the environment setup is part of the onboarding experience.

### Security

The committed `uv.lock` file means no user ever silently installs a different package version than the one that was tested. With the current setup, a `pip install --upgrade` during `rebuild_venv` can break scripts at any point if an upstream package introduces a breaking change. With a lockfile, updates are explicit, reviewed in PRs, and applied deliberately.

### Developer experience

Today a contributor who wants to run HyDE scripts locally has to read through `pip_env.py` to understand how to set up the environment. With uv it is a single command:

```bash
uv sync --project Configs/.local/lib/hyde
```

Adding a new dependency for a script is equally straightforward:

```bash
uv add <package> --project Configs/.local/lib/hyde
```

No custom infrastructure to understand.

### What we replace

Currently `pyutils/pip_env.py` (~380 lines) manually re-implements venv creation, integrity validation, multi-step package upgrades, and dynamic on-demand installs. All of this exists because pip and venv do not provide it out of the box.

uv provides all of it natively:

| Custom code today | uv equivalent |
|---|---|
| `python -m venv` + pip seed | handled automatically by `uv sync` |
| 3-step upgrade loop (install → list outdated → upgrade) | `uv sync` (single atomic operation) |
| `is_venv_valid()` — custom Python version check | uv validates and recreates the env automatically |
| `v_import()` / `v_install()` — dynamic on-demand installs | same public API, backed by `uv add` at dev time + `uv sync` at runtime |
| `requirements.txt` | `pyproject.toml` + `uv.lock` |

The result is ~300 fewer lines of custom environment management code that we no longer need to maintain.

### uv availability during installation

`restore_cfg.sh` must ensure uv is present before setting up the Python environment. The check will follow the same pattern used for other HyDE dependencies:

```bash
if ! command -v uv &>/dev/null; then
    print_log -warn "[uv]" "uv not found, installing..."
    if command -v pacman &>/dev/null; then
        sudo pacman -S --noconfirm uv
    else
        # Fallback: official installer (distro-agnostic)
        curl -LsSf https://astral.sh/uv/install.sh | sh
        source "$HOME/.local/bin/env"  # reload PATH for current session
    fi
fi
```

uv is available natively on Arch (`pacman -S uv`) and the official installer covers all other distros without requiring pip or cargo.

### Proposed changes

1. **`pyproject.toml`** — add `[project]` with all dependencies; becomes the single source of truth alongside a committed `uv.lock`
2. **`pip_env.py`** — rewrite internals to use `uv sync` and `uv add`; public API fully preserved (`v_import`, `v_install`, CLI subcommands) so all callers are unaffected
3. **`restore_cfg.sh`** — add uv availability check and install step before `pyinit`
4. **`hyde-shell`** — `run_pip()` redirected to `uv`; `python_activate()` and `python_initialized()` unchanged (uv manages standard PEP 405 venvs — `source bin/activate` still works)
5. **Shell scripts** — `gamelauncher.sh`, `gpuinfo.sh`: replace bare `python3` with explicit venv python path, removing the implicit `sys.path` self-repair dependency
6. **`requirements.txt`** — deprecated in this PR, removed in follow-on

---

### Outcome

HyDE's Python runtime becomes a reproducible, isolated module. Scripts behave consistently across machines, across Python upgrades, and regardless of what else the user has installed. First-run setup is faster. Dependencies are locked and auditable. Adding new Python dependencies becomes a one-liner. Environment-related bug reports become a category that largely disappears.

### Use Case

**Users on Arch (the primary target):** uv is in the official repos (`pacman -S uv`). The first `pyinit` after install goes from potentially 20–30 seconds to under 3. No interaction needed, no waiting.

**Users on other distros (Debian, Fedora, NixOS, openSUSE…):** Today the install can silently break if the system Python differs from what the venv was built with, or if pip resolves a different package version. With uv and a committed lockfile, every user gets the exact same packages regardless of distro.

**Users with complex Python setups (conda, pyenv, pipx, multiple venvs):** These are the users generating the hardest-to-reproduce bug reports. Right now, if they have an active conda env or a pyenv shim, HyDE scripts may pick up wrong packages or wrong Python versions. After this change, HyDE is fully blind to whatever else is on the machine.

**Contributors adding new Python functionality:** Instead of editing `requirements.txt` and hoping reviewers notice the version, a single `uv add <package>` updates both `pyproject.toml` and `uv.lock` atomically. The diff makes the intent obvious and the lockfile makes the change reviewable.

**Maintainers:** Roughly 300 fewer lines of custom environment management code to review, debug, and keep compatible with new Python releases. When Python 3.14 ships and something breaks, the fix is in uv — not in our `is_venv_valid()`.

### Proposed solution

The venv path stays at `$XDG_STATE_HOME/hyde/pip_env/` — no migration needed for existing installs. uv is pointed at that path via the `UV_PROJECT_ENVIRONMENT` variable, so it manages the same directory that `python_activate()` already knows how to source.

**`pyproject.toml`** becomes the single dependency declaration:

```toml
[project]
name = "hyde-pyutils"
version = "0.1.0"
requires-python = ">=3.11"
dependencies = [
    "loguru==0.7.3",
    "pulsectl==24.12.0",
    "requests==2.33.0",
    "inotify_simple",
    "pywayland",
    "PyGObject",
    "PyQt6",
]
```

**`pip_env.py`** is rewritten to a thin shell around uv. The public API (`v_import`, `v_install`, CLI subcommands) is preserved unchanged so every caller continues to work without modification. Internally:

- All `subprocess` calls to `python -m venv` and `pip` are replaced by `uv sync` and `uv add`
- The 3-step upgrade loop in `rebuild_venv()` collapses to a single `uv sync` call
- `is_venv_valid()` is simplified: the `import pip` check is removed (uv venvs don't ship pip), only the Python binary existence and version match from `pyvenv.cfg` are kept
- A `_get_uv()` helper resolves the uv binary and raises a clear error if missing
- Scattered `sys.path.insert` calls are consolidated into a single `_inject_site_packages()` helper

**`hyde-shell`** requires only one change: `run_pip()` currently pipes to the venv's `pip` binary, which uv venvs don't include. It is redirected to `uv pip --python <venv_path>`. `python_activate()` and `python_initialized()` are untouched — uv produces a standard `bin/activate` script.

---

#### First install

```text
restore_cfg.sh
  │
  ├─ check: command -v uv
  │    ├─ found → continue
  │    └─ not found → pacman -S uv  (or curl astral.sh/uv/install.sh)
  │
  └─ hyde-shell pyinit
       │
       └─ python_initialized()
            │
            └─ uv sync
                 --project  $LIB_DIR/hyde
                 --python    system python
                 --env       $XDG_STATE_HOME/hyde/pip_env/
                 │
                 reads  → pyproject.toml  (dependency list)
                 reads  → uv.lock         (pinned versions)
                 writes → pip_env/        (isolated venv, ~seconds)
```

#### Subsequent runs (normal use)

```text
shell script  (e.g. gamelauncher.sh, gpuinfo.sh)
  │
  └─ $XDG_STATE_HOME/hyde/pip_env/bin/python  script.py
       │
       runs against the locked venv — no activation needed,
       no system Python involved, no sys.path injection
```

```text
hyde-shell  <command>.py
  │
  └─ python_activate()
       │
       source  pip_env/bin/activate   ← standard PEP 405 script, unchanged
       │
  └─ exec python  <command>.py
```

#### Adding a new dependency (contributor workflow)

```text
uv add <package> --project Configs/.local/lib/hyde
  │
  updates → pyproject.toml
  updates → uv.lock
  │
commit both files → all users get the pinned version on next sync
```

#### Environment rebuild (after Python upgrade or broken env)

```text
hyde-shell pyinit   OR   python pip_env.py rebuild
  │
  └─ uv sync  (detects Python version mismatch → recreates venv automatically)
```

### Alternatives

_No response_

### Additional Information

_No response_

0xGeN02:
Closes #1683 

## Summary by CodeRabbit

* **New Features**
  * Switch to the `uv` tool for dependency management and automatic installation if missing.
  * Apps and launchers now run against a consistent bundled Python environment for more predictable behavior.

* **Chores**
  * Project dependencies declared centrally in project metadata.
  * Legacy per-component requirements removed; environment handling consolidated.

* **Refactor**
  * Environment management rewritten for more reliable install/sync and package operations.


kRHYME7:
@rubiin you guys are the python people. Thanks in advance!!!

Configs/.local/lib/hyde/pyproject.toml
```diff
@@ -1,3 +1,17 @@
...
+requires-python = ">=3.11"
+dependencies = [
+    "loguru==0.7.3",
+    "pulsectl==24.12.0",
+    "requests==2.33.0",
+    "inotify_simple",
+    "pywayland",
+    "PyGObject",
+    "PyQt6",
+]
```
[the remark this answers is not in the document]
0xGeN02: fixed

Scripts/restore_cfg.sh
```diff
@@ -266,6 +266,18 @@ echo ""
...
+if ! command -v uv &>/dev/null; then
+    print_log -warn "[uv]" "uv not found, installing..."
+    if command -v pacman &>/dev/null; then
+        sudo pacman -S --noconfirm uv
+    else
+        curl -LsSf https://astral.sh/uv/install.sh | sh
+        # shellcheck disable=SC1091
+        source "$HOME/.local/bin/env" 2>/dev/null || true
+    fi
+fi
```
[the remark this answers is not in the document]
0xGeN02: fixed

Configs/.local/lib/hyde/pyproject.toml
```diff
@@ -1,3 +1,18 @@
...
+requires-python = ">=3.11"
+dependencies = [
+    "loguru==0.7.3",
+    "pulsectl==24.12.0",
+    "requests==2.33.0",
+    "inotify_simple",
+    "pywayland",
+    "PyGObject",
+    "PyQt6",
+    "pyamdgpuinfo"
```
kRHYME7: >     "pyamdgpuinfo"

Actually I did not add it in the  requirements as pamd should only be triggered if users have amd It hsoudl just be installed dynamically.

Configs/.local/bin/hyde-shell
```diff
@@ -433,9 +433,28 @@ run_command() {
...
+		list)
+			UV_PROJECT_ENVIRONMENT="$venv_path" uv pip list --python "${venv_path}/bin/python"
+			;;
+		*)
+			echo "Usage: hyde-shell pip [install|uninstall|list] [packages...]"
+			echo "  install   <pkg>  → runs: uv add"
+			echo "  uninstall <pkg>  → runs: uv remove"
+			echo "  list             → list installed packages"
+			;;
+	esac
```
kRHYME7, line 457: Will it break something if we call it `hyde-shell uv` instead?

kRHYME7 (changes requested):
Thanks. Everything looks good but please avoid breaking the existing envs. Never tested it yet though.

rubiin:
> @rubiin you guys are the python people. Thanks in advance!!!

UV's core advantage over pip is its speed , almost 10-100x so i say its a great addition.

Configs/.local/lib/hyde/pyproject.toml
rubiin: could be removed I guess

Configs/.local/bin/hyde-shell
rubiin: guess it might trigger some issues if there are references. Should rename all the invocations to that as well

Configs/.local/lib/hyde/pyproject.toml
0xGeN02: Okey let me refactor this to install only if needed

Configs/.local/bin/hyde-shell
0xGeN02: I think not, let me have a look

Configs/.local/lib/hyde/pyproject.toml
0xGeN02: Now it is dynamicly installed via `v_import` function only for AMD mqachines

Configs/.local/bin/hyde-shell
0xGeN02: Now you can do the comands as follows: `hyde-shell uv install <dependency>` or `hyde-shell uv list`

Configs/.local/bin/hyde-shell
```diff
@@ -432,10 +432,29 @@ run_command() {
 	fi
 }
 
-run_pip() {
-	python_activate
+run_uv() {
+	local project_dir="${LIB_DIR}/hyde"
+	local venv_path="${XDG_STATE_HOME:-$HOME/.local/state}/hyde/pip_env"
```
kRHYME7: @0xGeN02 can we rename pip_env into something else, something that indicates a virtual env. Also this prevents clashing with old installation. 

python-env maybe?

kRHYME7 (changes requested):

Configs/.local/bin/hyde-shell
rubiin: Another thing, how are we installing UV ? does the user need to manually install it . I found no mention of the binary install. @0xGeN02

rubiin: And unlike pip, UV is not shipped alongside with python

0xGeN02: okey, because this will be one of all of the envs maanged by uv and the user may have 10 for example we could change the name to: `hyde-env` or `hyde-pyenv` With this we make it easier to the user to know what it is or what it does imo

0xGeN02: what name do you like?
@kRHYME7 @rubiin

kRHYME7: We always assumed that we are using `~/.local/state/hyde` therefore, imo we cna use `~/.local/state/hyde/python-env`.

0xGeN02: Okey let me update the path then. Also i am working on a refactorization to make the pyenv a class instead of methods so we can now make internal methods not exposable to the end user

rubiin: are we going to add the UV binary on the install script as well ? @kRHYME7

0xGeN02:
Finaly i dindt do a class because would be an overkill for this amount of code and it would be worst to apply the cli commands so I just refactorized and cleaned the functions for mor clarity

rubiin (changes requested):

0xGeN02:
image

image

0xGeN02:
Everything seems good now, just let me imporve the install/uninstall pipeline and its ready to merge

0xGeN02:
image
image

Do you like the install/uninstall animations of uv or shall I not stream it??

Configs/.local/bin/hyde-shell
0xGeN02: I think that having the binary it is a better idea than curl it but we may need to update the binary recurrently so there is a trade-off

0xGeN02:
image

I just added a passthroug also to make possible to use all the commands given by the uv tool from our script for example

`python-env uv cache clean` runs `uv cache clean` inside our own virtual enviroment

0xGeN02:
image

Now it is done. I think we dont need anything else for now to manage all the venv.

kRHYME7:
@0xGeN02 Will be merging this now. Please look for regressions. Thank you! 

BTW can we develeopers use the rc branch as our stable branch ? this way we can track bugs before hitting the master branch. 

It would be much better to use the dev branch too! Thank you!

Configs/.local/lib/hyde/amdgpu.py
kRHYME7, line 1: BTW , @0xGeN02  what happens if in user side they install package? eg the pyamdgpu lib?

will it still work with uv.lock?

0xGeN02: I added a fix for more stability, now works like this:

Optional dependencies are declared in `pyproject.toml` under `[project.optional-dependencies]` and installed on-demand via `v_import`:

```python
python_env.v_import("pyamdgpuinfo", extra="amd")
```

So now they are declared under the uv.lock and there will be no conflicts for the user after doing `git pull`

Give me some minutes to test everything

0xGeN02: When the module is missing, `v_import` runs `uv sync --extra <group>`, installing all packages in that group at once. If the module is already present, it imports directly with no side effects.

This keeps upstream `pyproject.toml` and `uv.lock` clean 

PD: user-specific packages are never added via uv add, avoiding merge conflicts on updates

0xGeN02: see #1687

PR #1687, opened 2026-04-07
feat: update pyamdgpuinfo integration with optional-dependencies

0xGeN02:
This continues #1684

## Summary by CodeRabbit

* **New Features**
  * Added optional AMD GPU support as an installable extra.

* **Bug Fixes**
  * Logger input validated; invalid levels now default to INFO.
  * Idle-inhibitor no longer closes the display connection when audio stops.

* **Chores**
  * Simplified dependency loading across utilities; several components now expect required modules to be preinstalled.
  * Adjusted installation/extra handling for AMD support.
  * Updated wiki submodule reference.

### DIFF
--- a/Configs/.local/lib/hyde/amdgpu.py
+++ b/Configs/.local/lib/hyde/amdgpu.py
@@ -9,10 +9,9 @@ sys.path.insert(0, os.path.dirname(os.path.abspath(__file__)))
 
 logger = logger.get_logger()
 
-python_env.v_import(
-    "pyamdgpuinfo"
-)  # fetches the module by name // does `pip install --update pyamdgpuinfo` under the hood
-import pyamdgpuinfo
+python_env.v_import("pyamdgpuinfo", extra="amd")
+
+import pyamdgpuinfo #noqa: E402
 
 
 def format_frequency(frequency_hz: int) -> str:
--- a/Configs/.local/lib/hyde/amdgpu.py
+++ b/Configs/.local/lib/hyde/amdgpu.py
@@ -1,17 +1,13 @@
 #!/usr/bin/env python
 import json
-import os
-import sys
 import pyutils.logger as logger
 import pyutils.python_env as python_env
-
-sys.path.insert(0, os.path.dirname(os.path.abspath(__file__)))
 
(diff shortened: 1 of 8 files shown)


0xGeN02:
@rubiin @kRHYME7 now we dont need to import the dependencies with `python_env.v_import(<dependency>)` because it is preinstalled and synced by default with *uv*.

We only need  `python_env.v_import(<dependency>, extra=<group>)`  with the **extra** dependencies because they are not preinstaled by default on the user's machine.

Also rubbin plz let me know if anything breake because i removed the sys.path() from all import sections. I tested the files and work correclty but idk if that afects other files or scripts.

0xGeN02:
### Before

```python

import pyutils.python_env as python_env

sys.path.insert(0, os.path.dirname(os.path.abspath(__file__)))
python_env.v_import("requests")
import requests  # noqa: E402

```

### After

```python
import requests
```

0xGeN02:
Also one last thing @kRHYME7 which is the python version that is needed for HyDE?? I think would be a nice idea to "force" all machines to install the same pytohn version so all HyDE users use same python version and same enviroment

kRHYME7:
latest version is fine for now, so yeah the current version.


BTW will pyamdgpu lib dynamically install if needed?

0xGeN02:
Yes, the amdgpuinfo lib will be in the uv.lock but only installed on the user's machine if the script is executed.
So it is dynamicly installed, all the `[project.optional-dependencies]` will work as amdgpuinfo, also you can create different families for example if we want to add support for nvidia devices we will create a new family of dependencies called nvidia=["cuda-python", "pynvidia"] will be installed only dinamicly for the nvidia devices via the v_import function after executing a script

0xGeN02:
Also the user can sync the dependency manualy via ```python-env uv --hyde sync --extra amd```

0xGeN02:
This is ready to merge. Also see the PR for the website with the new page explaining all about python_env commands and workflow

kRHYME7:
Please don't hesitate to open a PR Incase of any regressions. Thank you !

PR #1735, opened 2026-05-07
feat: optional matugen support as alternative dcol backend

### LINKED ISSUE #1701: [Feat] Optional Matugen support
FYI:

Matugen handles colors dynamically, and deploy templates. Only

Wallbash handles colors using the dcol(dominant colors) and deploys templates too. But also handles thumbnail generation.

Planning to strip those functions into. wallbash/dcol and wallbash/thumb
This way we can Add optional matugen support over wallbash/dcol. 

Matugen solves the old issues with dcol cache generation where we can't figure out why some users cannot extract colors using imagemagick. Matugen will become a safety net for those systems.

Todo:
See if Feasible

0xGeN02:
## Summary

Adds optional [matugen](https://github.com/InioX/matugen) (Material You color generation) support as an alternative color extraction backend for wallbash/dcol. This addresses the long-standing issue where some users' systems fail to extract colors via imagemagick.

Closes #1701

## What this does

- **New file: `wallbash-matugen.sh`** — Drop-in alternative to `wallbash.sh` that generates an identical 89-line `.dcol` file using matugen's tonal palettes instead of imagemagick's k-means clustering.
- **Modified: `wallpaper/cache.sh`** — Added `fn_generate_dcol()` dispatcher function that routes dcol generation to the appropriate backend based on user configuration.
- **Modified: `pkg_extra.lst`** — Added `matugen-bin` as an optional (commented) package.

## Configuration

Users can set `DCOL_BACKEND` in their Hyde config:

| Value | Behavior |
|---|---|
| *(unset)* / `imagemagick` | Default — original wallbash.sh, no change in behavior |
| `matugen` | Use matugen exclusively (auto-fallback to imagemagick if matugen not installed) |
| `auto` | Try imagemagick first, fallback to matugen if it fails or produces invalid output |

## How the mapping works

Matugen generates Material You tonal palettes (6 palettes × 18 tones). These are mapped to dcol's 4-group structure:

| dcol group | matugen palette | Purpose |
|---|---|---|
| Group 1 (pry1) | `neutral` | Background/surface colors (dark, desaturated) |
| Group 2 (pry2) | `primary` | Main accent color |
| Group 3 (pry3) | `secondary` | Muted accent |
| Group 4 (pry4) | `tertiary` | Complementary accent |

Each group maps 9 tones to `xa1`–`xa9` (dark→bright in dark mode, reversed in light mode).

## Why matugen?

1. **Safety net** — Some users have reported imagemagick failing to extract colors (broken kmeans on certain systems/images). Matugen uses its own Rust-based color extraction.
2. **Material You harmony** — Matugen produces perceptually uniform, harmonious color schemes by design (based on Google's HCT color space).
3. **Zero impact on default** — Without setting `DCOL_BACKEND`, behavior is 100% unchanged.

## Testing

- Tested with 5+ wallpapers, all produce exactly 89 lines with valid shell syntax
- Live-tested wallbash template deployment with matugen-generated colors
- Verified fallback behavior (matugen not found → imagemagick; imagemagick fails → matugen)

## Dependencies

- `matugen-bin` (AUR) — only required if user opts into matugen backend
- `python3` — already a HyDE dependency (used for JSON parsing in wallbash-matugen.sh)

## Summary by CodeRabbit

* **New Features**
  * Optional Material You–based palette backend and CLI wrapper that can produce an alternate color file and integrate with the app’s color selection, including automatic dark/light mode detection.
* **Bug Fixes**
  * Palette generation now validates outputs, warns on unexpected results, and reports success with details.
* **Chores**
  * Added a commented optional package entry for the new backend.


0xGeN02:
## Note on color differences between backends

**ImageMagick** and **matugen** produce different colors from the same wallpaper — this is by design:

### ImageMagick (wallbash.sh)
- Uses **k-means clustering** to extract 4 dominant pixel colors directly from the image
- Accents (xa1-xa9) are generated by varying brightness/saturation on the **exact hue** of each dominant color
- Result: colors that **match the wallpaper pixels**

### Matugen (wallbash-matugen.sh)
- Extracts **1 source color** (most saturated) from the wallpaper
- Generates 6 **harmonious tonal palettes** using Google's Material You algorithm (HCT color space)
- Result: a **perceptually uniform color scheme** derived from the wallpaper, not the literal pixels

### Example (same wallpaper):
```
ImageMagick: pry1=#130E13  (actual dark pixel from image)
Matugen:     pry1=#1E1A1D  (neutral tone 10, harmonically derived)
```

Matugen is not meant to replicate imagemagick's output — it's an alternative approach that guarantees valid color extraction on systems where imagemagick's kmeans fails, while producing aesthetically coherent palettes.

0xGeN02:
Closes #1701

kRHYME7:
wow!!! 🤯 You beat me to it!
Planning to do this and here it came free♥️. I have 2 planned approaches and looks like your implementation is seamless.


So just to be sure if I understand this. As wallbash uses magick to cache .dcol files. We are also implementing the same logic but using matugen right? I did not check the script's full logic but can we  coexist both? maybe we can Output a `hash.mcol`? (Material color, dcol means dominant colors. Help me with naming)

In our theming script we can just source it and whatever color exist wins or if backend was set try to use it.


And I said I planned 2 approaches. 1st is this PR's implementation. Second is having  a matugen template. This one's kinda complex because we need to resolve what template to use and explicitly switching them means we won't have a wallbash fallback, if matugen doesn't have the template. And more of this soon let's polish this matugen caching for now.

0xGeN02:
Thanks! Yes, you've got it exactly right.

**Current implementation**: `wallbash-matugen.sh` generates the same `.dcol` format (89 lines, same variable names) so it's a transparent drop-in. The dispatcher in `cache.sh` just decides which script to call.

**Your `.mcol` idea is better** — having both coexist:
- `hash.dcol` = imagemagick dominant colors (existing behavior)
- `hash.mcol` = matugen material colors (new)

Then the theming script (`color.set.sh`) can source based on backend preference:
```bash
# Pseudocode for color.set.sh sourcing logic:
if [[ "$DCOL_BACKEND" == "matugen" ]] && [[ -f "$mcol_file" ]]; then
    source "$mcol_file"
elif [[ -f "$dcol_file" ]]; then
    source "$dcol_file"
fi
```

This way:
- Both caches are always generated (no re-computation when switching backends)
- `dcol` remains the default/fallback
- Users can switch backends without waiting for cache regeneration

I'll update the PR to implement the `.mcol` coexistence approach. Will keep the same variable names inside (`dcol_pry1`, etc.) so downstream templates don't need changes — only the file extension and sourcing logic differ.

Re: approach #2 (matugen templates) — agreed, let's nail the caching layer first. The template approach would be a separate feature on top of this foundation.

0xGeN02:
opencode just auto answered you ahahhaha didnt notice was running still

0xGeN02:
Updated! The implementation now uses the `.mcol` coexistence approach:

### How it works now:
1. **`cache.sh`** always generates `.dcol` (imagemagick). If matugen is installed, it also generates `.mcol` alongside.
2. **`color.set.sh`** picks which file to source based on `DCOL_BACKEND`:
   - `unset`/`imagemagick` → sources `hash.dcol` (default, zero behavior change)
   - `matugen` → sources `hash.mcol` if it exists, falls back to `hash.dcol`

### Benefits:
- Both caches coexist (`~/.cache/hyde/dcols/hash.dcol` + `hash.mcol`)
- Switching backend is instant (no re-computation needed)
- `.dcol` is always the fallback safety net
- `.mcol` generation is opportunistic (only if matugen binary is present)

### Files changed:
- `wallbash-matugen.sh` → outputs `.mcol` extension
- `wallpaper/cache.sh` → `fn_generate_dcol()` generates both
- `color.set.sh` → sourcing logic picks `mcol` or `dcol` based on `DCOL_BACKEND`
- Also removed orphaned `Wiki` submodule reference (repo no longer exists)

0xGeN02:
Now I think it is done, any other comment on this?

kRHYME7:
Ohhhh I scanned through the code but isn't it better to use full python for the wallbash equivalent? This way it is easier to debug. Any reason for the design?

matugen.py is fine for this IMO.

0xGeN02:
The only reason was to mirror wallbash.sh

0xGeN02:
Added a command for `hyde-shell` to change the backend easier:

```bash
hyde-shell dcol.backend --get              # prints current backend
hyde-shell dcol.backend --set matugen      # switches to matugen
hyde-shell dcol.backend --set imagemagick  # switches back
hyde-shell dcol.backend --select           # rofi picker
```

kRHYME7:
Let's call it `hyde-shell color.backend` instead.

0xGeN02:
Done

kRHYME7:
Will be merging this after we setup lua soon. Also  this  chnages will not land the old master branch. We badly need to clean up our repo. 

I am planning to freeze the current `master` branch (user are using that )  and slowly move our development to `main` where we will do a full lua support for hyprland and remove all the .conf files. and removing all deprecated scripts. The lua support will force users to update the keybindings so it is safe to nuke the deprecated scripts.

PR #1808, opened 2026-07-09
fix: respect configured terminal for system updates

juliogc:
## Summary

- make `hyde-shell systemupdate` use the configured terminal instead of invoking Kitty directly
- support Ghostty's `-e` command argument and title syntax
- document the user-visible fix in the changelog

## Root cause

The system update script invoked `kitty` directly, bypassing HyDE's configured `TERMINAL` value.

## Validation

- `bash -n Configs/.local/lib/hyde/system.update.sh`
- `git diff --check`
- `pre-commit run --files Configs/.local/lib/hyde/system.update.sh CHANGELOG.md`

## Summary by CodeRabbit

* **Bug Fixes**
  * System updates now open in your configured terminal instead of a fixed default.
  * Improved support for Ghostty and other terminal apps when launching updates.

* **Documentation**
  * Updated release notes to reflect the terminal handling fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ All notable changes to 'HyDE' will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/), and this project adheres to _Modified_ [CalVer](https://calver.org/). See [Versioning](https://github.com/HyDE-Project/HyDE/blob/master/RELEASE_POLICY.md#versioning-yymq) For more info
  -->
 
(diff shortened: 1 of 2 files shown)


kRHYME7:
We use xdg-terminal-exec specs to do this. The wiring  is not yet in the master branch. Using this spec we can run `xdg-terminal-exec`  to launch our default terminal (with fallback)





image



With this approach, we can have a proper way to select the proper default terminal 

image

kRHYME7:
Thank you!

Issue #1824, opened 2026-07-21
[Bug]: Keybind Hint Initialization Failed


Shooter991as:
### Description

just updated hyde to the latest version by changing and pulling from latest branch and installed it. Almost everything is fine  but the MainMod + ? showing Keybind Hint Initialization failed

### Steps to Reproduce

update

### Expected Behavior

work

### Actual Behavior

dont

### Screenshots

<img width="1920" height="190" alt="Image" src="https://github.com/user-attachments/assets/4e2093cd-f2bf-4279-b46b-bb042967a5c8" />

### Environment

[hyprland_systeminfo.txt](https://github.com/user-attachments/files/30228105/hyprland_systeminfo.txt)

valkyrieAttr:
Most likely, this may be an issue with `hyprctl -j binds` or with the `hint-hyprland.py`.. It is parsing a wrong-malformed JSON that is hindering the process.

If your keybinds_hint.py is throwing this error:
```json.decoder.JSONDecodeError: Expecting value: line 15 column 16 (char 303)```

Then it may be possible that JSON-Parser failed. When `json.load()` tries to parse it, it chokes on a line e.g., 15, col 16. The script then likely catches that error, and then goes to sleep.

Can you please try and run `hyprctl -j binds` and look at the very top or bottom of the output?

[Malformed.json](https://github.com/user-attachments/files/30280068/malformed.json)

**Comment by valkyrieAttr** [comment]
I have confirmed that this fatal bug exists in Hyprland 0.56.0.

```
Hyprland 0.56.0 built from branch v0.56.0 at commit 36b2e0cfe0c6094dbc47bd42a437431315bb3087 clean (version: bump to 0.56.0).
Date: Mon Jul 20 07:50:48 2026
Tag: v0.56.0, commits: 7625

Libraries:
Hyprgraphics: built against 0.5.1, system has 0.5.1
Hyprutils: built against 0.14.0, system has 0.14.0
Hyprcursor: built against 0.1.13, system has 0.1.13
Hyprlang: built against 0.6.8, system has 0.6.8
Aquamarine: built against 0.13.0, system has 0.13.0

Version ABI string: 36b2e0cfe0c6094dbc47bd42a437431315bb3087_aq_0.13_hu_0.14_hg_0.5_hc_0.1_hlg_0.6
no flags were set
```
The above file breaks core JSON rules in multiple ways, meaning no JSON parser will be able to read it. 

[...]

PR #1826, opened 2026-07-23
fix: prevent fork loop and JSON hang in keybinds hint

snakeice:
## Problem

Two bugs in the keybinds hint system were causing system-wide performance degradation (fork storms with ~568 forks/s, load average spikes to 100+).

### Bug 1: `keybinds_hint.sh` — infinite self-re-execution (`exec $0`)

The script uses `exec $0` in two places as a fallback when rofi produces no output or when a dispatch fails. This causes the script to re-execute itself indefinitely, spawning hundreds of python3 processes per second.

**Before:**
```bash
RUN() {
    case ... in *"Not enough arguments"*) exec $0 ;; esac
}
...
else
    exec $0
fi
```

**Fix:** Replace `exec $0` with `exit 0`. If the script reaches a dead end, it should exit, not loop.

### Bug 2: `hint-hyprland.py` — infinite retry on broken JSON

`get_hyprctl_binds()` had a `while True` loop that retried `hyprctl binds -j` forever when JSON parsing failed. On Hyprland 0.56.0, `hyprctl binds -j` returns malformed JSON (fields shifted, non-numeric values in numeric fields), causing the script to hang indefinitely in the retry loop.

```
$ hyprctl binds -j | python3 -m json.tool
# JSONDecodeError: Expecting value: line 15 column 16 (char 303)
```

**Fix:**
- Try JSON first with a 5s timeout
- If JSON is invalid, fall back to parsing the text output of `hyprctl binds` (no `-j` flag)
- All subprocess calls have `timeout=5` to prevent indefinite hangs

## Impact

- Fork rate dropped from ~568/s to ~7/s
- System load returned to normal (was spiking to 100+ on a 14-thread CPU)
- Keybinds hint (`Super+/`) now works on Hyprland 0.56.0 where `hyprctl binds -j` is broken

## Testing

Tested on:
- Hyprland 0.56.0 (commit 36b2e0c)
- Kernel 7.1.4-1-cachyos
- Python 3.11

```bash
# Before: hangs forever
$ python3 hint-hyprland.py --format rofi
# (no output, CPU spins)

# After: works correctly
$ python3 hint-hyprland.py --format rofi | head -3
 Window Management  ::: ::: Window Management
SUPER + Q > close focused window ::: exec ::: hyde-shell dontkillsteam
ALT + F4 > close focused window ::: exec ::: hyde-shell dontkillsteam
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved keybinding detection by falling back to text-based parsing when structured output isn’t available.
  * Added bounded command execution for keybind retrieval to avoid indefinite waiting.
  * Prevented the keybind dispatch script from repeatedly restarting when dispatch arguments are missing or selection can’t be performed.
  * Enhanced normalization of keybinding details, including modifier, repeat, lock, mouse, release, and description handling.


Configs/.local/lib/hyde/keybinds/hint-hyprland.py
```diff
@@ -8,27 +8,69 @@
...
+        )
+        binds = []
+        current = {}
+        for line in result.stdout.strip().split("\n"):
+            line = line.strip()
+            if line == "bindd" or line == "bind":
+                if current:
+                    binds.append(current)
+                current = {}
+            elif ":" in line:
```
juliogc, line 40: Suggestion: this fallback is parsing a schema, rather than expressing distinct control flow. A field-to-parser mapping would make new Hyprland fields a one-line addition and, with explicit defaults, ensure partially parsed binds cannot later fail with missing keys in `expand_meta_data`.

```python
BIND_DEFAULTS = {
    "modmask": 0,
    "submap": "",
    "key": "",
    "keycode": 0,
    "catch_all": False,
    "description": "",
    "has_description": False,
    "dispatcher": "",
    "arg": "",
    "repeat": False,
    "locked": False,
    "mouse": False,
    "release": False,
    "non_consuming": False,
    "allow_input_capture": False,
}

parse_bool = lambda value: value.lower() == "true"
parse_int = lambda value: int(value) if value.isdigit() else 0
FIELD_PARSERS = {
    "modmask": ("modmask", parse_int),
    "submap": ("submap", str),
    "key": ("key", str),
    "keycode": ("keycode", parse_int),
    "catchall": ("catch_all", parse_bool),
    "description": ("description", str),
    "dispatcher": ("dispatcher", str),
    "arg": ("arg", str),
    "repeat": ("repeat", parse_bool),
    "locked": ("locked", parse_bool),
    "mouse": ("mouse", parse_bool),
    "release": ("release", parse_bool),
    "non_consuming": ("non_consuming", parse_bool),
    "allow_input_capture": ("allow_input_capture", parse_bool),
}

# `startswith` also accommodates other Hyprland bind variants.
if line.startswith("bind"):
    if current:
        current["has_description"] = bool(current["description"])
        binds.append(current)
    current = BIND_DEFAULTS.copy()
elif ":" in line:
    key, _, value = line.partition(":")
    field = FIELD_PARSERS.get(key.strip())
    if field:
        destination, parser = field
        current[destination] = parser(value.strip())
```

kRHYME7:
https://github.com/HyDE-Project/HyDE/issues/1824#issuecomment-5069498169

Configs/.local/lib/hyde/keybinds/hint-hyprland.py
snakeice: make sense, adjusted :smile:

kRHYME7:
Hi! I don't think adding a fix for a minor bug ( broken json) is needed. 0.56.1 fixes the json output already. Hyprland is fast changing, adding more guardrails for minor bug that was eventually fix doesn't make sense to me.

Thanks!

PR #1859, opened 2026-07-28
fix: expand variables in shader source includes

### LINKED ISSUE #1853: [Bug]: shaders.lua does not expand variables in the //!source= directive, wallbash shader fails to compile
### Environment

- HyDE `v26.7.4` (`c971ca63`)
- Hyprland 0.56.1, Arch Linux

### Problem

`Configs/.config/hypr/shaders/wallbash.frag` declares its colour source with an unexpanded variable:

```glsl
//!source=$XDG_CACHE_HOME/hyde/wallbash/colors.inc
```

`parse_source_include()` in `Configs/.local/lib/hyde/shaders.lua` takes the value verbatim:

```lua
local source = line:match("^%s*//%s*!source%s*=%s*(.-)%s*$")
if source and source ~= "" then
    if source:sub(1, 1) ~= "/" then
        source = shader_dir .. "/" .. source
    end
```

Because `$XDG_CACHE_HOME` is never expanded, the path becomes
`~/.config/hypr/shaders/$XDG_CACHE_HOME/hyde/wallbash/colors.inc`, which does not exist. The include is then dropped silently in `compile_shader()`, since the file is only appended when `lfs.attributes(source_include, "mode") == "file"`.

The predecessor `shaders.sh` went through the shell, where the variable expanded naturally, so this is a regression introduced by the Lua port.

### Result

`wallbash.inc` defines the colour by macro name only:

```glsl
#define WALLBASH_COLOR wallbash_pry1
```

and `colors.inc`, which defines `wallbash_pry1`, is missing from the compiled output. Hyprland refuses the shader and shows the error overlay on every start:

```
Screen shader parser: Error compiling shader: 0:56(25): error: `wallbash_pry1' undeclared
0:56(20): error: cannot construct `vec3' from a non-numeric data type
0:56(20): error: operands to arithmetic operators must be numeric
0:56(20): error: initializer of const variable `COLOR' must be a constant expression
0:79(19): warning: `COLOR' used uninitialized
```

### Reproduce

1. Select the `wallbash` shader.
2. Run `hyde-shell shaders --reload`.
3. `head -8 ~/.local/state/hyde/compiled.cache.glsl` — the `#define wallbash_pry1 …` block from `colors.inc` is absent.

### Expected

`//!source=` resolves `$VAR` and `${VAR}` before the path is used, with the XDG defaults as fallback when the variable is not exported. A missing include should also surface an error instead of being skipped silently.

RAprogramm:
## Description

Fixes #1853.

`wallbash.frag` points at its colour definitions with `//!source=$XDG_CACHE_HOME/hyde/wallbash/colors.inc`, but `parse_source_include()` took the value verbatim. The variable was never expanded, so the path resolved to `~/.config/hypr/shaders/$XDG_CACHE_HOME/hyde/wallbash/colors.inc`, the include was dropped, and Hyprland rejected the shader on every start with `` `wallbash_pry1' undeclared `` in the error overlay. The predecessor `shaders.sh` went through the shell, where the variable expanded on its own, so this arrived with the Lua port.

Changes:

- `Configs/.local/lib/hyde/shaders.lua` — expand `$VAR` and `${VAR}` in the `//!source=` value before resolving it, falling back to the XDG defaults when the variable is not exported.
- `Configs/.local/lib/hyde/shaders.lua` — a declared source include that cannot be found is now reported instead of silently skipped, so a broken shader fails loudly at compile time rather than at render time.
- `Configs/.local/lib/hyde/luautils/xdg.lua` — a variable that is set but empty is treated as unset. The base directory specification calls for this, and without it an exported `XDG_CACHE_HOME=` produced an empty prefix instead of the default.

## Type of change

## Checklist

## Testing

Expansion was covered case by case, run twice: once with `XDG_CACHE_HOME`, `XDG_CONFIG_HOME` and `XDG_STATE_HOME` exported empty, once with them unset. Both runs pass identically.

| Input | Result |
| --- | --- |
| `$XDG_CACHE_HOME/hyde/wallbash/colors.inc` | `~/.cache/hyde/wallbash/colors.inc` |
| `${XDG_CACHE_HOME}/x.inc` | `~/.cache/x.inc` |
| `$HOME/a/$XDG_STATE_HOME` | both expanded in one pass |
| `$NOSUCHVAR/tail` | `/tail`, unknown variable collapses to empty |
| `/absolute/path.inc` | unchanged |
| `relative.inc` | unchanged, still resolved against the shader directory |

Integration: `lua Configs/.local/lib/hyde/shaders.lua --reload` on Hyprland 0.56.1 now emits `#define wallbash_pry1 72,56,40` from `colors.inc` into `compiled.cache.glsl`, the shader compiles, and the error overlay is gone. `hyprctl configerrors` is empty afterwards.

Negative path: pointing `//!source=` at a non-existent file now fails with `compile shader: source include not found: /home/<user>/.cache/hyde/wallbash/NOSUCH.inc`, quoting the expanded path. `compiled.cache.glsl` is left untouched in that case, since the check runs before the output file is opened.

`luac -p` is clean on both changed files.

## Summary by CodeRabbit

* **Bug Fixes**
  * Correctly treat empty environment variables as unset, using appropriate fallback paths for XDG-related directories.
  * Shader include directives now expand environment variables consistently (including fallbacks).
  * Shader compilation now fails with a clear error when an expanded include path exists but isn’t a file.
* **Enhancements**
  * Improved compatibility with XDG-style environment configuration and environment-driven shader setup.


RAprogramm:
Addressed in 972827c.

`dirs()` now applies the same nil-or-empty check as `env()`, so an exported but empty `XDG_DATA_DIRS` or `XDG_CONFIG_DIRS` falls back to its default instead of collapsing the list.

The difference is visible directly: with `XDG_DATA_DIRS=` exported, the previous version returns a single element, the home directory prepended by the `home` argument, while the search paths `/usr/local/share` and `/usr/share` are lost. After the change the same environment returns all three entries in order.

Covered case by case against both an empty and an unset environment, with identical results in each run: `data`, `config`, `cache` and `state` resolve under `$HOME`, `dirs.data` is `$XDG_DATA_HOME`, `/usr/local/share`, `/usr/share`, and `dirs.config[2]` is `/etc/xdg`. The shader path this PR is about still compiles afterwards and `compiled.cache.glsl` still carries the colour definitions from `colors.inc`.

PR #1860, opened 2026-07-28
fix: resolve lua before shell and retire superseded scripts

### LINKED ISSUE #1854: [Bug]: Scripts removed in the Lua release survive the upgrade and shadow their replacements
### Environment

- HyDE `v26.7.4` (`c971ca63`), upgraded from `v26.4.x`
- Hyprland 0.56.1, Arch Linux

### Problem

The Lua release replaced a large number of shell scripts with Lua equivalents (`shaders.sh` → `shaders.lua`, `animations.sh` → `animations.lua`, `workflows.sh` → `workflows.lua`, `window.pin.sh` → `window.pin.lua`, `open.sh` → `open.lua`, `batterynotify.sh` → `batterynotify.lua`, `color/dconf.sh` → `color/dconf.lua`, `system.update.sh` → `system.update.py`, and others).

After upgrading, the removed `.sh` files are still present in `~/.local/lib/hyde`, even though the `hyde` dot declares pruning:

```toml
[[hyde.files]]
clean_target = true
action = "sync"
source_root = "Configs/.local/lib"
target_root = "${HOME}/.local/lib"
paths = "hyde"
```

Comparing the deployed directory against the tree at `origin/master` shows 26 files that no longer exist upstream but survive locally, `shaders.sh`, `animations.sh`, `workflows.sh`, `themeselect.sh`, `wbarconfgen.sh`, `swwwallpaper.sh` among them.

This matters because `Configs/.local/bin/hyde-shell` resolves `.sh` **before** `.lua`:

```sh
if [[ -f "$dir/${1}.sh" ]]; then
    exec bash "$dir/${1}.sh" "${@:2}"
elif [[ -f "$dir/${1}.lua" ]]; then
```

So every leftover script shadows its replacement.

### Reproduce

On a system upgraded from a pre-Lua release:

```
$ hyde-shell shaders --help
Usage: /home/<user>/.local/lib/hyde/shaders.sh [OPTIONS]
```

The legacy script answers instead of `shaders.lua`. Several of these scripts call helpers that no longer exist, so the commands behind the affected keybindings fail.

### Expected

Either the deployment actually removes files that disappeared upstream, or `hyde-shell` prefers `.lua` over `.sh` when both are present. Preferably both, so an interrupted or partial deployment cannot leave the shell dispatching to stale code.

RAprogramm:
## Description

Fixes #1854.

After upgrading from a pre-Lua release, the shell helpers that were replaced by Lua and Python equivalents stay on disk, and `hyde-shell` resolved `.sh` before `.lua`, so the leftovers answered instead of the scripts that replaced them. On an upgraded machine `hyde-shell shaders --help` printed the usage of `shaders.sh`, and 25 further scripts were shadowed the same way. Several of them call helpers that no longer exist, so the affected keybindings fail.

Changes:

- `Configs/.local/bin/hyde-shell` — resolve `.lua` before `.sh`. `.py` and the bare executable keep their existing places in the chain, and a command that only ships as `.sh` is unaffected.
- `Scripts/migrations/v26.7.4.sh` — new migration that clears the superseded scripts out of `~/.local/lib/hyde` on existing installs. Nothing is deleted: each file is moved to `$XDG_STATE_HOME/hyde/migration/v26.7.4`, keeping its relative path, so a local modification can still be recovered. The name sorts last, which is what `install.sh` picks up.

The deployment side of the report, `clean_target = true` not removing files that disappeared upstream, lives in `deez-dots` and is not addressed here. The migration covers the installs that are already in that state, and the resolution order means a leftover cannot shadow a replacement even if one slips through later.

## Type of change

## Checklist

## Testing

`bash -n` is clean on both files. `find Scripts/migrations -maxdepth 1 -type f -printf '%f\n' | sort -r | head -n 1` returns `v26.7.4.sh`, so `install.sh` selects the new migration.

Resolution order, checked against a stub script directory:

| Present in the directory | Executed |
| --- | --- |
| `probe.lua` and `probe.sh` | `probe.lua` |
| `probe.sh` only | `probe.sh` |
| `probe.py` only | `probe.py` |
| `probe`, executable | `probe` |

Migration, checked in a sandboxed `HOME` and `XDG_STATE_HOME`:

- Seeded with `shaders.sh`, `animations.sh`, `workflows.sh`, `window.pin.sh`, `color/dconf.sh`, plus `shaders.lua` and an unrelated `custom.sh`. All five superseded files moved, the nested path preserved as `hyde/migration/v26.7.4/color/dconf.sh`, and `shaders.lua` and `custom.sh` untouched.
- A second run is a no-op and exits 0.
- A missing `~/.local/lib/hyde` exits 0 without output.

The list of superseded files was derived from `git diff --name-status <last pre-Lua commit> dev -- Configs/.local/lib/hyde`, filtered to the entries that are still deployed by another dot. `screenshot/grimblast`, `app2unit` and `xdg-terminal-exec` arrive as blobs from `Scripts/dots/hyde.toml` and are deliberately not in the list; `hyde-config` is resolved from `PATH` and shadows nothing, so it is left alone as well.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved command resolution so Lua implementations are selected before shell implementations when both are available.
  * Prevented outdated helper scripts from overriding newer command behavior.

* **Maintenance**
  * Added a deployment migration that backs up superseded helper/config files to XDG state storage (with a fallback path).
  * Added clearer move reporting by summarizing relocated files, and failing the migration when moves or directory creation encounter errors.

### DIFF
--- a/Configs/.local/bin/hyde-shell
+++ b/Configs/.local/bin/hyde-shell
@@ -294,10 +294,10 @@ run_command() {
         [[ -d "$dir" ]] && SCRIPT_DIRS+=("$dir")
     done
 
+    # Lua is resolved before shell, so a script left behind by an older release
+    # cannot shadow the implementation that replaced it.
     for dir in "${SCRIPT_DIRS[@]}"; do
-        if [[ -f "$dir/${1}.sh" ]]; then
-            exec bash "$dir/${1}.sh" "${@:2}"
-        elif [[ -f "$dir/${1}.lua" ]]; then
+        if [[ -f "$dir/${1}.lua" ]]; then
             l_env="${XDG_STATE_HOME:-$HOME/.local/state}/hyde/lua_env"
             if [[ ! -d "$l_env" ]]; then
                 print_log -warn "Lua environment not found!" " Please run 'hyde-shell luainit' to set up the environment."
@@ -315,6 +315,8 @@ run_command() {
             export LUA_PATH="$l_env/share/lua/$lua_ver/?.lua;$l_env/share/lua/$lua_ver/?/init.lua;$LUA_PATH"
             export LUA_CPATH="$l_env/lib/lua/$lua_ver/?.so;$LUA_CPATH"
             exec lua "$dir/${1}.lua" "${@:2}"
+        elif [[ -f "$dir/${1}.sh" ]]; then
+            exec bash "$dir/${1}.sh" "${@:2}"
         elif [[ -f "$dir/${1}.py" ]]; then
             exec "${XDG_STATE_HOME:-$HOME/.local/state}/hyde/python_env/bin/python" "$dir/${1}.py" "${@:2}"
         elif [[ -f "$dir/${1}" && -x "$dir/${1}" ]]; then
--- a/Scripts/migrations/v26.7.28.sh
+++ b/Scripts/migrations/v26.7.28.sh
@@ -1,0 +1,60 @@
+#!/usr/bin/env bash
+
+# The Lua release replaced a set of shell helpers with Lua and Python
+# equivalents. Deployment overwrites files but does not delete the ones that
+# disappeared upstream, and hyde-shell used to resolve ".sh" before ".lua", so
+# the leftovers kept answering in place of the scripts that replaced them.
+#
+# Nothing is deleted here. Anything found is moved aside so it can be restored
+# if a local change is still needed.
+
+lib_dir="${HOME}/.local/lib/hyde"
+backup_dir="${XDG_STATE_HOME:-${HOME}/.local/state}/hyde/migration/v26.7.28"
+
+superseded="
+animations.sh
+app2unit.sh
+batterynotify.sh
+color/dconf.sh
+dontkillsteam.sh
+hypr.altab.lua
+hypr.unbind.sh
+open.sh
+parse.config.py
+parse.json.py
+shaders.sh
+swwwallbash.sh
+swwwallcache.sh
+swwwallpaper.sh
+swwwallselect.sh
+sysmonlaunch.sh
+system.update.sh
+systemupdate.sh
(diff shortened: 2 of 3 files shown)


RAprogramm:
Both migration findings are addressed in b2243a8.

An occupied backup destination is now reported and left untouched instead of being overwritten, and `mkdir`, `mv` and collision failures are each counted and reported, with a non-zero exit when anything was left in place. `install.sh` already treats a failing migration as a warning rather than a hard stop, so the new exit status surfaces the problem without aborting the restore.

Verified in a sandboxed `HOME` and `XDG_STATE_HOME`:

| Case | Result |
| --- | --- |
| first run | 3 files moved, exit 0 |
| source restored, second run | `skipped shaders.sh, a backup already exists at …`, exit 1, the earlier backup keeps its original content |
| backup directory not writable | `failed to create the backup directory for shaders.sh`, exit 1, source left in place |
| nothing to move | no output, exit 0 |

On the resolution order I did not move `.py` ahead of `.sh`, because that would change behaviour upstream deliberately ships. Scanning `Configs/.local/lib/hyde` on `dev` for basenames carrying more than one of `.lua`, `.sh`, `.py` returns exactly one collision, `pm.sh` and `pm.py`, and `pm.py` opens with "Python rewrite of the pm.sh helper". Both are shipped, `.sh` currently wins, and reordering would silently swap the package manager wrapper for every user.

The concern itself is real but narrower than it looks: the only superseded `.sh` whose replacement is a `.py` is `system.update.sh`, and the migration in this PR moves it out of the way. `.lua` is safe to promote because no shipped basename carries both `.lua` and `.sh`, so nothing upstream depends on the old precedence there.

kRHYME7 (approved):

kRHYME7:
thanks!

PR #1877, opened 2026-07-29
fix: double slurp grimblast

zeerayne:
# Pull Request

## Description

Update `grimblast` to latest version to [fix double slurp](https://github.com/hyprwm/contrib/commit/3dcbce715ae8b93107fa8632db15bf976862a573)

## Type of change

- [x] **Bug fix** (non-breaking change which fixes an issue)
- [ ] **New feature** (non-breaking change which adds functionality)
- [ ] **Breaking change** (fix or feature that would cause existing functionality to not work as expected)
- [ ] **Documentation update** (non-breaking change; modified files are limited to the documentations)
- [ ] **Technical debt** (a code change that does not fix a bug or add a feature but makes something clearer for devs)
- [ ] **Other** (provide details below)

## Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/HyDE-Project/HyDE/blob/master/CONTRIBUTING.md) document.
- [x] My code follows the code style of this project.
- [x] My commit message follows the [commit guidelines](https://github.com/HyDE-Project/HyDE/blob/master/COMMIT_MESSAGE_GUIDELINES.md).
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added a changelog entry.
- [ ] I have added necessary comments/documentation to my code.
- [ ] I have added tests to cover my changes.
- [x] I have tested my code locally and it works as expected.
- [ ] All new and existing tests passed.

## Summary by CodeRabbit

- **New Features**
  - Added support for multiple screenshot output formats, including appropriate clipboard MIME types.
  - Added an option to open the parent directory after saving a screenshot.
  - Improved area selection behavior and handling when selection is canceled.

- **Bug Fixes**
  - Fixed an issue causing double slurp behavior.
  - Updated Grimblast for more reliable screenshot and cleanup handling.

PR #1775, opened 2026-06-06
feat(i18n): add gettext zh_CN runtime POC

WeissHymmnos:
## Summary

This PR adds a small gettext-based runtime i18n proof of concept for HyDE, focused on Simplified Chinese (`zh_CN`). It replaces the earlier custom JSON language-pack direction with a more standard gettext-style workflow.

The scope is intentionally small: this PR establishes the foundation and demonstrates it on notifications and the keybinding hint UI only.

## Changes

- Add gettext translation sources under `Source/locale/`:
  - `hyde.pot`
  - `zh_CN/LC_MESSAGES/hyde.po`
  - `build.py`
  - `README.md`
- Track generated runtime artifacts:
  - `Configs/.local/share/hyde/locale/zh_CN/LC_MESSAGES/hyde.mo`
  - `Configs/.local/share/hyde/locale/zh_CN.sh`
- Add a Python gettext helper:
  - `Configs/.local/lib/hyde/pyutils/i18n.py`
- Upgrade the existing Bash localization POC:
  - `Configs/.local/lib/hyde/shutils/l10n.sh`
- Localize a small set of high-impact runtime strings in:
  - `Configs/.local/lib/hyde/notifications.py`
  - `Configs/.local/lib/hyde/keybinds_hint.sh`

## Design notes

- `.po` files are the editable translation source.
- `.mo` files are generated for Python/gettext consumers.
- Bash uses a lightweight generated associative-array map from the same `.po` file.
- Language resolution order is `HYDE_LANG`, `I18N_LANGUAGE`, `LC_MESSAGES`, `LANG`, then English/original strings.
- Missing translation files and missing keys fall back to the original English strings.
- This PR does not localize the installer, Waybar-wide strings, or the rest of the repository.

## Test plan

Passed locally:

```bash
python3 Source/locale/build.py
msgfmt --check --output-file=/tmp/hyde-zh_CN-check.mo Source/locale/zh_CN/LC_MESSAGES/hyde.po
python3 -m py_compile Source/locale/build.py Configs/.local/lib/hyde/pyutils/i18n.py Configs/.local/lib/hyde/notifications.py
bash -n Configs/.local/lib/hyde/shutils/l10n.sh Configs/.local/lib/hyde/keybinds_hint.sh Configs/.local/share/hyde/locale/zh_CN.sh
[...]

PR #1890, opened 2026-07-30
#1889 fix: point shipped callers at files that still exist

### LINKED ISSUE #1889: [Bug]: shipped scripts and the config registry still point at files the Lua release deleted
> Correction to the first version of this text: it also claimed `screenshot.sh` pointed at a missing helper. That was wrong — `Scripts/dots/hyde.toml:97-103` declares a separate `[grimblast]` dot that downloads the helper from hyprwm/contrib straight into `~/.local/lib/hyde/screenshot/`, so the path in the script is correct. Everything below has since been verified against every dot, including the remote `blob+` ones and `post_command` hooks.

Sibling of #1854: that issue covers the leftover scripts, `#1860` fixed resolution order and `Scripts/migrations/v26.7.4.sh` retires the leftovers. Several items here are the second-order effect of that cleanup — the migration removes the very orphan these call sites depended on, turning a working line into a broken one.

## Hard failures

**`hyde-shell themeswitch` does not exist.** `themeswitch.sh` was deleted by `6eb07e2d` and `Scripts/migrations/v26.7.4.sh:35` lists it as superseded, so the leftover is actively removed. `hyde-shell` has no alias table — `run_command` tries only `$dir/$1.lua`, `.sh`, `.py`, then an executable — so the command hard-fails with `Command not found: themeswitch`. Four shipped call sites:

| File | Line |
| --- | --- |
| `Configs/.local/share/waybar/modules/custom-theme.jsonc` | 6-7 |
| `Configs/.local/share/waybar/modules/custom-hyde-menu.jsonc` | 14-15 |
| `Configs/.local/share/waybar/layouts/macos.jsonc` | 97 |
| `Configs/.local/lib/hyde/wallpaper.kon.sh` | 45 (`$scrDir/themeswitch.sh`) |

All of `modules/` and `layouts/` ship via `Scripts/dots/waybar.toml:24-28`. `wallpaper.kon.sh` is reachable from Dolphin: `Scripts/dots/dolphin.toml:11-19` ships `hydewallpaper.desktop` with `Exec=wallpaper.kon.sh`. The replacement is `theme.switch.sh`, which `hyde-shell` already advertises as `theme.switch`.

**The config editor creates dead files.** `Configs/.local/share/hyde/config-registry.toml:71-89` offers `hypr/userprefs.conf`, `hypr/keybindings.conf`, `hypr/windowrules.conf` and `hypr/nvidia.conf`. Nothing loads any of them: `Configs/.local/share/hypr/hyde.lua` requires only Lua modules and user overrides live in `${XDG_CONFIG_HOME}/hypr/hyprland.lua` (shipped `preserve` by `Scripts/dots/hyprland.toml:23-27`). `hydectl` flags a missing entry in its list, but `config.EditConfigFile` runs `ExpandPath` → `os.MkdirAll` → `$EDITOR` without ever stat'ing the target, so accepting one creates the parent directory and, on save, a config file Hyprland never reads. `[hyprland.files.main]` should point at `${XDG_CONFIG_HOME}/hypr/hyprland.lua`.

## Wrong environment variable, two places

The shipped Lua entry point is `${XDG_DATA_HOME}/hypr/hyde.lua` (`Scripts/dots/hyprland.toml:11-16`, which explicitly ignores `*.conf`).

| Source | Value | State |
| --- | --- | --- |
| `Configs/.local/lib/hyde/shell/activate:130`, `activate.fish:115` | `hyde.lua`, then the two `/usr/...` fallbacks | correct |
| `Configs/.config/fish/conf.d/hyde.fish:77` | `${XDG_DATA_HOME}/hypr/hyprland.conf` | deleted by `6eb07e2d` |
| `Configs/.config/uwsm/env-hyprland.d/00-hyde.sh:28` | `${XDG_DATA_HOME}/hypr/hyprland.lua` | shipped by no dot |

Both are scoped, not fatal: `00-hyde.sh` sources `activate` at lines 15-26 before its own default, and `activate`/`activate.fish` reset `HYPRLAND_CONFIG` to empty and then set it, so they win over anything fish exported. The fish line therefore bites only a session that never activates — the case in discussion #1849, where the reporter quotes this exact snippet and the maintainer answers `start-hyprland -- -c $XDG_DATA_HOME/hypr/hyde.lua`. It also poisons the `"${HYPRLAND_CONFIG##*.}" == "lua"` branches in `theme.switch.sh:126-141`, `color.set.sh:7` and `color/hypr.sh:208`. PR #1873 already carries the right shape for a candidate list.

## Reachable but silently degraded

**`Configs/.local/lib/hyde/system.monitor.sh:63`**, reachable from `CTRL + SHIFT + ESCAPE` (`Configs/.local/share/hypr/lua/key_binds.lua:61`):

```
term=$(grep -E '^\s*'"$term" "$HOME/.config/hypr/keybindings.conf" | cut -d '=' -f2 | xargs)
```

Two defects. The file is gone. And `$term` is first assigned on that very line, so it is unset on the right-hand side and the pattern degrades to a bare `^\s*` — the grep was already meaningless before the deletion. The documented fallback does not exist either: `SYSMONITOR_TERMINAL` appears only in the script's own `show_help`; `[sysmonitor]` in `Configs/.local/share/hyde/schema/config.toml:214-217` and `schema/schema.toml:773-788` declares just `execute` and `commands`. So the chain is `$TERMINAL` or nothing, and with nothing the monitor is launched with no terminal attached.

**`Configs/.local/lib/hyde/globalcontrol.sh:335-340`** and its Lua port `Configs/.local/lib/hyde/luautils/theme/parser.lua:134-141` grep defaults from six paths, of which `${XDG_DATA_HOME}/hyde/hyde.conf` and `${XDG_DATA_HOME}/hyde/hyprland.conf` are shipped by no dot (`hyde.conf` went in `b89f0e92`, `hyprland.conf` in `6eb07e2d`); the four `/usr/...` copies could only come from a distro package. `get_hyprConf` is exported and used by roughly twenty scripts, so this is not a break: the primary source `$HYDE_THEME_DIR/hypr.theme` still answers, and where it does not, callers chain to an empty value — rofi and dunst get an empty font rather than an error.

## Dead code

**`Configs/.local/lib/hyde/hyde-launch.sh:6`** runs `${LIB_DIR}/hyde/open.sh`, deleted by `6eb07e2d` and listed at `Scripts/migrations/v26.7.4.sh:22`. Nothing in the repository invokes `hyde-launch` — a full grep finds only its own deprecation notice — so it ships purely because `hyde.toml` syncs the whole directory. Fix is cosmetic, or delete the file.

**`Configs/.local/lib/hyde/keybinds_hint.sh:6-7`** seeds `kb_hint_conf` with three deleted `.conf` paths, and the variable is never read again; `ROFI_KEYBIND_HINT_CONFIG` appears nowhere else in the tree. The hint reads live binds through `keybinds/hint-hyprland.py`, so the whole assignment is dead. Note #1826 and #1775 also edit this file.

## Unrelated but adjacent

`Configs/.local/lib/hyde/grimblast` is referenced by nothing — neither by path nor bare name. `screenshot.sh:109,126,145` and `screenrecord.sh:83` both invoke `$LIB_DIR/hyde/screenshot/grimblast`, the vanilla upstream copy the `[grimblast]` dot downloads. So the in-repo file is dead code, and the double-selection fix in #1877 lands on a script nothing runs — the live bug is in the downloaded vanilla one.

## Expected

Shipped code references files that exist. Where a reference was to something the Lua release retired, it points at the replacement or goes away.

RAprogramm:
Closes #1889

## Hard failures

**`hyde-shell themeswitch` did not exist.** `themeswitch.sh` was deleted by `6eb07e2d` and `Scripts/migrations/v26.7.4.sh:35` retires the leftover, and `hyde-shell` has no alias table — `run_command` tries `.lua`, `.sh`, `.py`, executable — so every call hard-failed. Repointed at `theme.switch` in `Configs/.local/share/waybar/modules/custom-theme.jsonc`, `modules/custom-hyde-menu.jsonc`, `layouts/macos.jsonc`, and at `theme.switch.sh` in `Configs/.local/lib/hyde/wallpaper.kon.sh` (reachable from Dolphin's "Set As Wallpaper", shipped by `Scripts/dots/dolphin.toml:11-19`).

**The config editor created dead files.** `config-registry.toml` offered `hypr/{userprefs,keybindings,windowrules,nvidia}.conf`, none of which anything loads. `hydectl` flags a missing entry in the list, but `config.EditConfigFile` runs `ExpandPath` → `os.MkdirAll` → `$EDITOR` without stat'ing the target, so accepting one created the directory and, on save, a file Hyprland never reads. The four entries collapse into one pointing at `${XDG_CONFIG_HOME}/hypr/hyprland.lua`, which is where user overrides actually belong (`Scripts/dots/hyprland.toml:23-27`, `preserve`).

## Environment variable

The deployed entry point is `${XDG_DATA_HOME}/hypr/hyde.lua` (`Scripts/dots/hyprland.toml:11-16`). `Configs/.config/fish/conf.d/hyde.fish:77` exported a `.conf` deleted by the Lua release, and `Configs/.config/uwsm/env-hyprland.d/00-hyde.sh:28` defaulted to `hypr/hyprland.lua`, which no dot deploys. Fish now walks the same candidate list as `shell/activate:49-61`, and uwsm defaults to `hyde.lua`.

Both were scoped rather than fatal — `00-hyde.sh` sources `activate` first, and `activate` resets `HYPRLAND_CONFIG` before setting it — so this bites the session that never activates, the case in discussion #1849, where the reporter quotes the fish line and the maintainer answers `-c $XDG_DATA_HOME/hypr/hyde.lua`. It also fed the `"${HYPRLAND_CONFIG##*.}" == "lua"` branches in `theme.switch.sh:126-141`, `color.set.sh:7` and `color/hypr.sh:208`.

## System monitor

`system.monitor.sh:63`, reachable from `CTRL + SHIFT + ESCAPE`, read the terminal out of the deleted `~/.config/hypr/keybindings.conf` — and `$term` was first assigned on that very line, so the pattern was a bare `^\s*` and the grep never meant anything even before the deletion. Removed. The documented fallback did not exist either: `SYSMONITOR_TERMINAL` appeared only in the script's `show_help`, so `terminal` is now declared in `schema/config.toml` and `schema/schema.toml`. With neither that nor `TERMINAL` set the script warns and moves to the next candidate instead of launching a console monitor with no terminal attached.

## Dead code

`hyde-launch.sh:6` called the deleted `open.sh`; it now calls `hyde-shell open`. Nothing in the tree invokes `hyde-launch` — a full grep finds only its own deprecation notice — so this is cosmetic; deleting the file instead would be equally correct.

`keybinds_hint.sh:6-7` seeded `kb_hint_conf` from three deleted `.conf` paths and never read the variable again; `ROFI_KEYBIND_HINT_CONFIG` appears nowhere else. The hint reads live binds via `keybinds/hint-hyprland.py`, so the assignment and the two variables feeding it are gone. Note #1826 and #1775 also touch this file; the removed lines are not the ones they edit.

## Deliberately not touched

`globalcontrol.sh:335-340` and `luautils/theme/parser.lua:134-141` still list `${XDG_DATA_HOME}/hyde/{hyde,hyprland}.conf` among six default-config paths. Those two are no longer deployed, but the primary source `$HYDE_THEME_DIR/hypr.theme` still answers and the six-path grep is a last resort that now returns empty rather than failing — and a user may legitimately keep their own copy there. Left for a separate decision.

`Configs/.local/lib/hyde/grimblast` is referenced by nothing: `screenshot.sh` and `screenrecord.sh` both invoke `$LIB_DIR/hyde/screenshot/grimblast`, the vanilla copy the `[grimblast]` dot downloads. Worth noting because #1877 fixes the double selection in the in-repo file, which nothing runs — the live bug is in the downloaded script.

## Tests

`tests/test_references.sh` (new) fails on the pre-change tree with 14 failures and passes after: it rejects any shipped config referencing `hyde-shell themeswitch`, `themeswitch.sh`, `hyde/open.sh` or the four retired `hypr/*.conf` paths, asserts the Lua entry point exists, and requires every shell integration that sets `HYPRLAND_CONFIG` to point at `hypr/hyde.lua`. The legacy `Scripts/restore_cfg.*` pair is out of scope, documented in the test.

Full suite: 9 cases, 0 failed.

## Summary by CodeRabbit

* **New Features**
  * Added a configurable terminal setting for console system monitoring (`sysmonitor.terminal`), with documented fallback behavior.

* **Bug Fixes**
  * Updated theme switching across Waybar (menus/modules) and Dolphin/Wallpaper workflows to use the current `theme.switch` actions.
  * Improved Hyprland Lua config detection and ensured the config editor serves the real Lua files.
  * Fixed Fish sessions started outside UWSM failing due to removed/incorrect Lua config paths.
  * Prevented the system monitor from launching when no terminal is configured.

* **Docs**
  * Added an upgrade guide for migrating from the Lua-based hyprlang setup and refreshed keybinding reference links.

* **Tests**
  * Added a reference-check test to catch retired config filenames and validate Lua entry-point/config-path expectations.


kRHYME7 (approved):

kRHYME7 (approved):

Configs/.local/lib/hyde/hyde-launch.sh
kRHYME7, line 1: Maybe we can nuke this file then

kRHYME7 (changes requested):

kRHYME7 (approved):

kRHYME7:
let's deprecate those with deprecation notices. deez-dots tracks those files so usually they are removed.

Issue #1897, opened 2026-07-30
[Bug]: HyDE doesn't work when initialized from TTY after migrating to Lua configs


time-slayer:
### Description

After updating to the new Lua configurations and trying to initialize the HyDE environment with `start-hyprland`, Hyprland throws a Lua timeout error

### Steps to Reproduce

1. Run `HyDE/Scripts/install.sh` or `install.sh -r` (tried both on non-fresh installl)
2. The installation finishes successfully to the end
3. Reboot
4. Log in via TTY
5. Start Hyprland with `start-hyprland -- -c ~/.local/share/hypr/hyde.lua`

### Expected Behavior

Hyprland successfully parses Lua configurations, launching into a fully functional HyDE environmen

### Actual Behavior

Instead of dropping into a configured Hyprland session, the screen displays a red error box with the following message:
```
Your config has errors:
require("dynamic"): ~/.local/share/hypr/lua/hyde/config.lua:98: Failed to parse TOML: ~/.local/lib/hyde/luautils/toml.lua:1259: [Lua] execution timed out in config reload
require("key_binds"): ~/.local/share/hypr/lua/hyde/binds.lua:53: [Lua] execution timed out in config reload
```

### Environment

- Main Hardware
- Arch Linux
- HyDE `master` (fd70502f95142c242d0ce2d2e569f6fd1d7dd626) and `dev` (362d50de862e44e4085a3225456c0b0335044a98)
- Hyprland 0.56.1

### Additional Information

`deez dots --filetree hyde`:
```
hyde by the_hyde_project
  `-- ~
      |-- .config
      |   |-- Kvantum
      |   |   |-- kvantum.kvconfig
      |   |   `-- wallbash
      |   |       |-- wallbash.kvconfig
      |   |       `-- wallbash.svg
      |   |-- gtk-3.0/settings.ini
      |   |-- hyde
      |   |   |-- config.toml
      |   |   `-- wallbash
      |   |       |-- README.md
      |   |       |-- always
      |   |       |   |-- cava.dcol
      |   |       |   |-- chrome.dcol
      |   |       |   |-- discord.dcol
      |   |       |   |-- spotify.dcol
      |   |       |   |-- vim.dcol
      |   |       |   `-- wayle.dcol
      |   |       |-- scripts
      |   |       |   |-- cava.sh
      |   |       |   |-- chrome.sh
      |   |       |   |-- code.sh
      |   |       |   |-- discord.sh
[...]

PR #1914, opened 2026-08-01
#1913 fix: load the session configuration without forking

### LINKED ISSUE #1913: [Bug]: the Hyprland configuration forks while loading and spends the reload budget before it runs
### Description

The shipped Hyprland configuration runs subprocesses while Hyprland is loading it, and on some machines that alone exhausts the reload budget and takes the session down. It is the mechanism behind #1897, where the reported timeouts land in code that is provably fast.

Hyprland executes the whole configuration in a single guarded call with a fixed budget (`src/config/lua/ConfigManager.cpp:746`, v0.56.1):

```cpp
if (guardedPCall(0, 0, 1, LUA_TIMEOUT_CONFIG_RELOAD_MS, "config reload") != LUA_OK) {
```

```cpp
static constexpr int LUA_WATCHDOG_INSTRUCTION_INTERVAL = 10000;
static constexpr int LUA_TIMEOUT_CONFIG_RELOAD_MS      = 1500;
```

Two properties follow, and both matter here. The 1500 ms covers the entire chunk — every `require`, not each one separately. And the watchdog is installed with `lua_sethook(..., LUA_MASKCOUNT, ...)`, so it counts VM instructions: it cannot fire while Lua sits blocked inside a C call, and it fires at the first Lua-heavy stretch after the deadline has already passed. The file and line in the error therefore name where execution happened to be, not what was slow.

The configuration blocks in two places:

| File | Call | Cost |
| --- | --- | --- |
| `Configs/.local/share/hypr/lua/env.lua:18` | `os.execute("nvidia-smi >/dev/null 2>&1")` | seconds on a hybrid laptop whose discrete GPU has to be woken to answer |
| `Configs/.local/share/hypr/lua/hyde/path.lua:84` | `io.popen("[ -d … ]")`, up to six per reload | a shell fork per directory probe |

### Steps to Reproduce

1. Boot a laptop with an NVIDIA discrete GPU alongside an integrated one, driver loaded, GPU idle.
2. Start the session: `start-hyprland -- -c ~/.local/share/hypr/hyde.lua`.

### Actual Behavior

The error overlay reports a timeout in whichever module follows the blocking call:

```
require("dynamic"): ~/.local/share/hypr/lua/hyde/config.lua:98: Failed to parse TOML: ~/.local/lib/hyde/luautils/toml.lua:1259: [Lua] execution timed out in config reload
require("key_binds"): ~/.local/share/hypr/lua/hyde/binds.lua:53: [Lua] execution timed out in config reload
```

That is exactly what #1897 shows, and the placement is the confirmation rather than a coincidence. `hyde.lua` requires `env` before `dynamic` and `key_binds`, and those two are the ones that fail while everything before them loads. `toml.lua:1259` is the UTF-8 scan over a 284-byte file, and `hyde/binds.lua:53` is the bind canonicaliser — a tight Lua loop is simply the first place a count hook can fire. The reporter's log shows the hardware the probe is expensive on:

```
drm: Found 2 GPUs
drm: Starting backend for /dev/dri/card1, with driver amdgpu
drm: Starting backend for /dev/dri/card0, with driver nvidia-drm with primary /dev/dri/card1
```

### Expected Behavior

Loading the configuration does not leave the process. Driver detection and directory probes are answered from `/proc` and `/sys` and from `io.open`, so the reload budget is spent on the configuration itself.

### Additional Information

Dropping `nvidia-smi` costs one thing worth stating: it also proved the userspace library matched the loaded kernel module. After the change, a machine whose driver was upgraded without a reboot gets the NVIDIA variables set where it previously did not — a session that is already broken either way, and the variables are inert without a working driver.

A regression test belongs with the fix: the shipped Hyprland Lua must not call `io.popen` or `os.execute`, and that is cheap to check.

RAprogramm:
# Pull Request

## Description

Fixes #1913. Explains and resolves the failure reported in #1897.

Hyprland executes the whole Lua configuration in one guarded call with a single 1500 ms budget (`ConfigManager.cpp:746`, `LUA_TIMEOUT_CONFIG_RELOAD_MS`), and installs its watchdog with `LUA_MASKCOUNT`, so the deadline is only ever noticed between VM instructions. Time spent blocked inside a C call is therefore unaccounted and uninterruptible, and the timeout is raised later, in whatever module happens to be executing plain Lua when the count hook next fires.

The configuration was blocking in two places:

| File | Call | Cost |
| --- | --- | --- |
| `hypr/lua/env.lua` | `os.execute("nvidia-smi >/dev/null 2>&1")` | seconds on a hybrid laptop that has to wake a sleeping discrete GPU to answer |
| `hypr/lua/hyde/path.lua` | `io.popen("[ -d … ]")`, up to six per reload | a shell fork per directory probe |

Both are now answered without leaving the process:

- NVIDIA detection reads `/proc/driver/nvidia/version`, and rules out a driver that is still coming up or on its way out by reading `/sys/module/nvidia/initstate`. When that file is absent the `/proc` entry alone decides, since the module only publishes it once registered.
- The directory probe opens the path and reads one byte: a directory fails with `EISDIR`, a regular file hands the byte over, a missing path does not open. The handle is closed on every outcome. `shell_quote` goes with the shell it existed for.

`tests/test_config_reload.sh` is added as the regression guard: no shipped Hyprland Lua file may call `io.popen` or `os.execute`. It was checked in both directions — 47 files pass today, and a file containing `io.popen` makes it fail.

The detection logic was exercised against a synthetic `/proc` and `/sys` for all four states: driver live → true, `initstate` reporting `going` → false, `initstate` absent → true, no `/proc` entry → false.

## Type of change

## Checklist

## Additional context

One property is deliberately given up: `nvidia-smi` also proved the userspace library matched the loaded kernel module. A machine whose driver was upgraded without a reboot now gets the NVIDIA variables set where it previously did not. That session is already broken by the mismatch itself, and the variables are inert without a working driver, so the trade buys back a budget the configuration cannot afford to spend.

Worth stating for #1897: this removes what consumed the budget, and the reporter's own numbers already ruled out the code the error named — the `config.toml` in question is 284 bytes, and the scan the timeout was reported at costs 0.0 ms at that size.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved NVIDIA GPU detection for more reliable discrete GPU session configuration.
  - NVIDIA-specific settings are now applied only when the driver is confirmed active.
  - Improved directory detection for paths containing spaces or quotes.
  - Directory checks safely reject invalid or non-directory paths without hanging.

- **Tests**
  - Added safeguards against unsafe subprocess execution in configuration scripts.
  - Expanded directory detection coverage, including command-injection protection and FIFO handling.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 - Hyprland: dropped the legacy hyprlang dot, the files it deployed no longer exist
 
 ### Fixed
+- Hyprland: a session on a machine with a discrete NVIDIA GPU no longer comes up with a timed-out configuration; driver detection and the directory probes read `/proc` and `/sys` instead of starting a subprocess, so the budget Hyprland allows the whole configuration is not spent waiting on one
 - Core: Pyprland commands keep their arguments when `nc`, `socat`, and `ncat` are unavailable, so commands such as `hyde-shell pypr toggle console` work through the CLI fallback again
 - Hyprland: a session started without `XDG_CONFIG_HOME`, such as one launched from a TTY, no longer dies with a Lua error before the first window; the unset variable is treated as unset instead of being pasted into a path
 - Hyprland: a home directory containing an apostrophe no longer makes the session load its libraries from the system directory instead of the user's own, and an empty `XDG_RUNTIME_DIR` reads as unset rather than resolving against the working directory
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 - Hyprland: dropped the legacy hyprlang dot, the files it deployed no longer exist
 
 ### Fixed
-- Hyprland: a session on a machine with a discrete NVIDIA GPU no longer comes up with a timed-out configuration; driver detection and the directory probes read `/proc` and `/sys` instead of starting a subprocess, so the budget Hyprland allows the whole configuration is not spent waiting on one
+- Hyprland: a session on a machine with a discrete NVIDIA GPU no longer comes up with a timed-out configuration; driver detection reads `/proc` and `/sys`, and the library directories are found by opening the candidate paths, so the budget Hyprland allows the whole configuration is no longer spent waiting on a subprocess
 - Core: Pyprland commands keep their arguments when `nc`, `socat`, and `ncat` are unavailable, so commands such as `hyde-shell pypr toggle console` work through the CLI fallback again
 - Hyprland: a session started without `XDG_CONFIG_HOME`, such as one launched from a TTY, no longer dies with a Lua error before the first window; the unset variable is treated as unset instead of being pasted into a path
 - Hyprland: a home directory containing an apostrophe no longer makes the session load its libraries from the system directory instead of the user's own, and an empty `XDG_RUNTIME_DIR` reads as unset rather than resolving against the working directory
--- a/Configs/.local/share/hypr/lua/env.lua
+++ b/Configs/.local/share/hypr/lua/env.lua
@@ -9,14 +9,30 @@ hl.env("DCONF_PROFILE",  ((os.getenv("XDG_CONFIG_HOME") ~= "" and os.getenv("XDG
 -- This is only a bare minimum to get NVIDIA working.
 -- User may need to add specific variables for their
 -- setup in ~/.config/hypr/hyprland.lua
+-- The probe reads two files and never leaves the process. It used to run
+-- nvidia-smi, which is what took whole sessions down: Hyprland gives the
+-- entire configuration 1500 ms and its watchdog counts VM instructions, so it
+-- cannot interrupt a blocked C call. A laptop that has to wake a sleeping
+-- discrete GPU to answer spends the budget here, and every module loaded after
+-- this one dies with a timeout reported somewhere unrelated.
 local function has_nvidia_working()
-	local f = io.open("/proc/driver/nvidia/version", "r")
-	if not f then
+	local version = io.open("/proc/driver/nvidia/version", "r")
+	if not version then
 		return false
 	end
-	f:close()
-	local ok, _, code = os.execute("nvidia-smi >/dev/null 2>&1")
-	return ok == true or code == 0
+	version:close()
+
+	-- The module registers /proc as it comes up, so a driver that is still
+	-- loading or on its way out is ruled out by its recorded state.
+	local initstate = io.open("/sys/module/nvidia/initstate", "r")
+	if not initstate then
+		return true
+	end
+
+	local state = initstate:read("*l")
+	initstate:close()
+
+	return state == "live"
 end
 if has_nvidia_working() then
 	hl.env("LIBVA_DRIVER_NAME", "nvidia")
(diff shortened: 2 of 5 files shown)


RAprogramm:
All three points from the review were valid and are fixed in `0af6f369` and `a86a9fa2`.

**The directory probe could hang.** Opening the bare candidate path meant a FIFO in place of a directory would block until a writer showed up, and this resolver runs before the session has a window on screen — a worse failure than the fork it replaced. Measured, not assumed: with a FIFO at `HOME/.local/lib`, the previous revision of the probe never returned and was killed at the timeout.

The probe now asks for the entry only a directory has:

```lua
local handle = io.open(path .. "/.", "r")
```

`dir/.` opens, `file/.` and `fifo/.` fail with `ENOTDIR` before anything is opened, so nothing can block. Verified against a directory, a symlink to one, a regular file, a FIFO, a character device and a missing path — true only for the first two, and no case takes measurable time. `tests/test_paths.sh` carries the FIFO case as the guard; it passes here and fails against the revision that opened the bare path.

**The scan missed the entry point.** `Configs/.local/share/hypr/hyde.lua` sits outside the `lua/` subdirectory and is the file Hyprland actually executes, so the root is now `Configs/.local/share/hypr` — 48 files instead of 47. The loop reads the file list line by line rather than splitting a command substitution, and a `grep` exit above 1 is reported as a failed scan instead of passing as "nothing found".

The scan deliberately stops at the shipped hypr tree. `Configs/.local/lib/hyde/luautils` is reached from the configuration too, but the same modules are used by command line tools where a subprocess is legitimate, and a blanket ban there would be wrong.

**The changelog conflated the two mechanisms.** Corrected: driver detection reads `/proc` and `/sys`, the library directories are found by opening the candidate paths.

Full suite green — 13 cases, 0 failed.

kRHYME7:
conflicts. 

you got write access now. so once I approve the PR, feel free to merge it when ready

RAprogramm:
Follow-up in `bf13e95a`, from re-reading the probe against what the shell one used to answer.

`test -d` reports true for a directory that grants search but not read; opening `dir/.` there fails with `EACCES`, so the previous revision of this branch skipped such a candidate and fell back to the system path. That is a behaviour change nobody asked for: a directory like that still serves files by name, which is the only thing `package.path` does with it. A refusal on permission grounds now counts as a directory, and the case is in `tests/test_paths.sh` — it fails against the revision without it, and it degrades to asserting the ordinary answer when the suite runs as root.

Two more points of coverage, same commit:

- A regular file standing in for a candidate directory has to be rejected too. That is the FIFO case without the hazard, and it is now asserted rather than assumed.
- The fork scan takes in `luautils/toml.lua`. It is the one module outside the shipped hypr tree that the configuration loads — `hyde/config.lua` requires it, and it requires nothing further — so it runs inside the same budget. The rest of `luautils` is shared with command line tools, where a subprocess is legitimate, and stays out of scope deliberately.

49 files scanned, full suite green at 13 cases.

kRHYME7 (approved):

kRHYME7:
resolve also bot before merging.

RAprogramm:
Last point from the review is fixed in `7049d345`.

It was right: `resolve()` printed `error: …` and still exited 0, so the two cases that assert a path is *not* chosen would have been satisfied by the module never loading at all. The helper now exits non-zero on a raise, the FIFO case distinguishes the three outcomes it can have — resolved, hung at the timeout, raised — and the regular-file case fails on a raise instead of reading it as a rejection.

Checked by pointing the suite at a `path.lua` that does nothing but raise: nine assertions fail, where the negative ones used to pass.

Thank you for the write access. #1914 and #1915 are green and mergeable; I will merge them once you are happy with this last round.

PR #1915, opened 2026-08-01
#1904 fix: declare the lgi build dependency and skip an unbuildable optional rock

### LINKED ISSUE #1904: [Bug]: ./install.sh "Failed to create LUA Environment"
### Description

./install.sh on fresh minimal arch throws an error "Failed to create LUA Environment"

I tried to install packages "lua" "luarocks" as recommended by users on official discord but no luck .. same issue 

### Steps to Reproduce

1. Install a minimal Arch with git base-devel
2. git clone --depth 1 https://github.com/HyDE-Project/HyDE ~/HyDE
3. cd ~/HyDE/Scripts
4. ./install.sh

### Expected Behavior

Arch with HyDE installed :)

### Actual Behavior

error "Failed to create LUA Environment"

### Screenshots

Image

### Environment

.

### Additional Information

_No response_

RAprogramm:
# Pull Request

## Description

Fixes #1904.

The install fails at the Lua step with `Failed to create Lua environment`. The output in the report names the cause: `luarocks` builds `lgi` from the upstream rockspec, and the build stops at

```
pkg-config --exists 'gobject-introspection-1.0 >= 0.10.8' --print-errors
Package 'gobject-introspection-1.0' not found
make[1]: *** [Makefile:55: .depcheck] Error 1
Error: Build error: Failed building.
```

Two separate defects meet there.

The build inputs are not declared. `Scripts/dots/deps.toml` lists `lua` and `luarocks`, and nothing that provides `gobject-introspection-1.0.pc` — which is why the dependency step reports "nothing to install" and the Lua step then fails on a system that has every declared package. On Arch that file and the matching headers live in `libgirepository`, which is now declared.

A single rock also decides the fate of the whole install. `install_rock` raises, `create_env` lets it through, and `install.sh` exits, so a machine that cannot build one optional binding never gets its dotfiles — even though `lgi` is already treated as special elsewhere in the same config through `snapshot_exclude`. Bootstrap entries can now mark themselves optional; an optional rock that fails is reported, named in a summary with the command to retry, and skipped, while a required one still stops the run.

`tests/test_lua_env.sh` covers all three statements, driving the module against a stub `luarocks`: an optional failure must not end `create`, it has to be reported, a required failure must still end it, and the shipped bootstrap list has to mark `lgi` optional. Checked in both directions — against this branch it passes, and against the current `dev` module it fails with exactly the traceback from the report.

## Type of change

## Checklist

## Additional context

Only the Arch list is touched, since that is the platform the report and the package name apply to. The equivalent on Debian and Fedora is `libgirepository1.0-dev` and `gobject-introspection-devel`; if the other package lists should carry it too, say so and it goes in.

A machine that skips `lgi` loses what depends on it — the battery notifier, the dconf colour writer and the notification helper — and keeps everything else. That is the trade the skip is for: a missing optional binding is a degraded desktop, an aborted install is none.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved installation reliability when optional Lua packages, including `lgi`, cannot be built.
  * Failed optional packages are reported and skipped while deployment continues.
  * Required package failures still stop installation as expected.
  * Added support for packages requiring introspection headers.

* **Documentation**
  * Added a changelog entry describing the updated installation behavior.

* **Tests**
  * Added coverage for optional and required package installation failures during environment setup and synchronization.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 - Hyprland: dropped the legacy hyprlang dot, the files it deployed no longer exist
 
 ### Fixed
+- Core: an install no longer ends at the Lua step on a machine that cannot build `lgi`; the introspection headers it compiles against are declared as a dependency, and an optional rock that still fails is reported and skipped instead of taking the run down before any dotfile is deployed
 - Core: Pyprland commands keep their arguments when `nc`, `socat`, and `ncat` are unavailable, so commands such as `hyde-shell pypr toggle console` work through the CLI fallback again
 - Hyprland: a session started without `XDG_CONFIG_HOME`, such as one launched from a TTY, no longer dies with a Lua error before the first window; the unset variable is treated as unset instead of being pasted into a path
 - Hyprland: a home directory containing an apostrophe no longer makes the session load its libraries from the system directory instead of the user's own, and an empty `XDG_RUNTIME_DIR` reads as unset rather than resolving against the working directory
--- a/Configs/.local/lib/hyde/pyutils/lua_env.json
+++ b/Configs/.local/lib/hyde/pyutils/lua_env.json
@@ -5,7 +5,7 @@
     {"name": "dkjson", "version": "2.11-1"},
     {"name": "luasocket", "version": "3.1.0-1"},
     {"name": "luafilesystem", "version": "1.9.0-1"},
-    "https://raw.githubusercontent.com/lgi-devs/lgi/master/lgi-scm-1.rockspec"
+    {"name": "https://raw.githubusercontent.com/lgi-devs/lgi/master/lgi-scm-1.rockspec", "optional": true}
   ],
   "snapshot_exclude": [
     {"name": "luasec", "version": "1.3.2-1"},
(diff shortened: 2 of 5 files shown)


kRHYME7:
conflicts

RAprogramm:
Follow-up in `42094f0a`, extending the tests to the path a user hits more often than a fresh install.

`hyde-shell luainit` refreshes the environment through `sync`, not `create`, and falls back to a full `rebuild` when `sync` fails. Without the same tolerance in `sync`, an unbuildable optional rock would tear the rocks tree down and rebuild it on every routine refresh — the original defect wearing different clothes. `sync` goes through the same helper as `create`, and `tests/test_lua_env.sh` now asserts both directions there: an optional failure is survived, a required one still ends the run.

Five cases in that file now. Checked against the module on `dev` as well, where three of them fail — including the `sync` case.

kRHYME7 (approved):

RAprogramm:
The nitpick from the review is fixed in `af70e72c`.

It was right that matching the package name proved little: the failing command line carries the name too, so a stub that merely fails could satisfy it. Both optional cases now look for the words only the handler writes — the skip, the summary of what was skipped, and the command that retries it — and both required cases assert that the run ended on the failed package rather than on any non-zero status.

Checked against the module on `dev`, where eight of those assertions fail.

Thank you for the write access. Green and mergeable here; I will merge once you are happy with this round.

PR #1923, opened 2026-08-03
fix: retire the hyprlang configuration the Lua release left behind

RAprogramm:
The Lua release deleted the hyprlang configuration from the repository, but deployment overwrites files and never removes the ones that disappeared upstream. The whole chain is still on every upgraded machine, and it is not inert — Hyprland falls back to `$XDG_CONFIG_HOME/hypr/hyprland.conf` whenever no Lua config is found, and that file sources the rest.

That is what @realronak is running in #1878. Three of the four symptoms reported there come from it and nothing else:

| Symptom | Where it comes from |
| --- | --- |
| `Executable not found: batterynotify.sh` | `$XDG_DATA_HOME/hypr/variables.conf` line 41 still names the shell script the `v26.7.4` migration moved aside |
| `SUPER + Q` and `ALT + F4` do nothing | `~/.local/share/hyde/keybindings.conf` lines 33-34 still bind them to `hyde-shell dontkillsteam` |
| `hyprctl systeminfo` reports `configProvider: hyprlang` | `~/.config/hypr/hyprland.conf` is what the session loaded |

`v26.7.4` moved the superseded *scripts* aside. Nothing moved the *configuration* that names them, which is why the scripts are gone and the references are not.

## What the migration does

Everything below is moved into `$XDG_STATE_HOME/hyde/migration/v26.8.2`, nothing is deleted — for most of these files that backup is the only copy of settings the user wrote.

| Root | Moved |
| --- | --- |
| `$XDG_CONFIG_HOME/hypr` | `animations.conf`, `hyprland.conf`, `keybindings.conf`, `monitors.conf`, `nvidia.conf`, `shaders.conf`, `userprefs.conf`, `windowrules.conf`, `workflows.conf`, `animations/`, `workflows/` |
| `$XDG_DATA_HOME/hypr` | `defaults.conf`, `dynamic.conf`, `env.conf`, `finale.conf`, `hyprland.conf`, `migration.conf`, `startup.conf`, `variables.conf`, `windowrules.conf` |
| `$XDG_DATA_HOME/hyde` | `hyprland.conf`, `keybindings.conf`, `templates/hypr/` |

The compositor configuration only. `hyprlock.conf`, `hypridle.conf` and `hyprsunset.conf` are still hyprlang and still shipped, as is `hypr/shaders/`.

## The gate

It runs only when the Lua entry point is in place: `$XDG_DATA_HOME/hypr/hyde.lua` readable, and `$XDG_CONFIG_HOME/hypr/hyprland.lua` carrying the loader from #1922. Taking the fallback away while nothing else can be found would leave Hyprland to generate a fresh default config, which is worse than the leftovers. Readable rather than merely present, because the loader opens it.

Depends on #1922 — without the loader the gate never opens.

## Tests

`tests/test_hyprlang_leftovers.sh` builds a machine in the shape described above and checks:

- every retired path is moved and recoverable from the backup
- `hyprlock.conf`, `hypridle.conf`, `hyprsunset.conf`, `hypr/shaders/`, the shipped `hypr/hyprlock.conf`, the Lua config and the entry point are untouched
- a second run does nothing
- no entry point, an unreadable entry point, or a user config that does not load HyDE yet: the fallback stays
- a retired path left as a dangling symlink is moved rather than skipped

Checked in both directions: the case fails against `dev`, and against a copy with the gate removed and the dangling-symlink test dropped it fails on exactly those three points.

Suite: 15 cases, 0 failed. `shellcheck --severity=error` clean.

Part of #1878.

## Summary by CodeRabbit

* **New Features**
  * Added an automated migration for obsolete Hyprland configuration files.
  * Preserves existing backups and safely handles dangling symbolic links.
  * Reports incomplete migrations and returns a failure status when items cannot be moved.

* **Bug Fixes**
  * Prevents migration from running when required Lua configuration is missing, unreadable, or inactive.
  * Ensures shipped configuration and Lua files remain intact.
  * Supports safe, repeatable migration runs without overwriting backups.

### DIFF
--- a/Scripts/migrations/v26.8.2.sh
+++ b/Scripts/migrations/v26.8.2.sh
@@ -1,0 +1,124 @@
+#!/usr/bin/env sh
+
+# The Lua release deleted the hyprlang configuration from the repository, but
+# deployment overwrites files and never removes the ones that disappeared
+# upstream. The whole chain is therefore still on every upgraded machine, and
+# it is not inert: Hyprland falls back to "$XDG_CONFIG_HOME/hypr/hyprland.conf"
+# whenever no Lua config is found, and that file sources the rest. A session
+# that lands on it runs the pre-Lua configuration — the battery notification
+# still names a script that was moved aside, and the close-window binds still
+# call one that no longer exists.
+#
+# Nothing is deleted here. Everything is moved into a backup directory, which
+# for most of these files is the only copy of settings the user wrote.
+#
+# This runs only when the Lua entry point is in place. Moving hyprland.conf out
+# of the way while nothing else can be found would leave Hyprland to generate a
+# fresh default config, which is worse than the leftovers.
+
+config_home="${XDG_CONFIG_HOME:-${HOME}/.config}"
+data_home="${XDG_DATA_HOME:-${HOME}/.local/share}"
+backup_dir="${XDG_STATE_HOME:-${HOME}/.local/state}/hyde/migration/v26.8.2"
+
+entry_point="${data_home}/hypr/hyde.lua"
+user_config="${config_home}/hypr/hyprland.lua"
+
+# Readable, not merely present: the loader opens this file, and an unreadable
+# one would fail there with the fallback configuration already moved away.
+if [ ! -r "${entry_point}" ]; then
+    echo "  ${entry_point} is missing or unreadable, leaving the hyprlang configuration in place"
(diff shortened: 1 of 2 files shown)


kRHYME7 (approved):

kRHYME7 (approved):
merge when ready

PR #1929, opened 2026-08-03
#1897 fix: resolve the shipped modules from the entry point's own path

RAprogramm:
Closes the remaining half of #1897.

Hyprland builds the Lua search path from the directory of the config it was started with and prepends nothing else — [`Lua::CConfigManager`](https://github.com/hyprwm/Hyprland/blob/main/src/config/lua/ConfigManager.cpp) does `configDir = path(m_mainConfigPath).parent_path()` and puts `configDir/?.lua;configDir/?/init.lua` in front of `package.path`. Which files a bare `require()` can see therefore depends entirely on which file the session was pointed at.

HyDE is reachable through two of them, and since #1922 both are live:

| Entry point | Search path anchored at | `require("lua.hyde.path")` |
| --- | --- | --- |
| `~/.local/share/hypr/hyde.lua`, named by `HYPRLAND_CONFIG` | `~/.local/share/hypr` | resolves |
| `~/.config/hypr/hyprland.lua`, what Hyprland picks on its own | `~/.config/hypr` | not found |

The second one is what a session started from a TTY or a display manager gets, and it ends the way it was reported in #1897:

```
Emergency mode tripped: A lua config error resulted in no binds being registered.
/home/timeslayer/.local/share/hypr/hyde.lua:17: module 'lua.hyde.path' not found:
no file '/home/timeslayer/.config/hypr/lua/hyde/path.lua'
```

The resolver is what every later path is built from, so it is the one module that may not be reached through a search path it has not set up yet. It now comes off the entry point's own location, which `debug.getinfo` reports for both entry points because Hyprland loads the config with `luaL_loadfile`. The module is registered under the name the rest of the tree uses, so requiring `hyde.path` later returns the same table rather than a second copy. That location also joins `pkg_paths` as the last entry, which makes a system-wide install find its own tree when `XDG_DATA_HOME` exists but holds no HyDE.

Nothing about the modules loaded after it changes: they still load by name, from the search path this block sets.

## Tests

`tests/test_lua_entry_point.sh` already covered the loader and its migration against a stub. It now also loads the real tree the way Hyprland loads it, from both entry points, and checks that the bootstrap completes, that every `hyde.path` field is resolved, that the resolver is registered once, and that the shipped tree is on the search path.

Against this branch:

```
test_lua_entry_point
  ok
```

Against `dev`, the first entry point reproduces the report verbatim:

```
fail: /tmp/.../.config/hypr/hyprland.lua stopped before the first require:
      module 'lua.hyde.path' not found: no file '/tmp/.../.config/hypr/lua/hyde/path.lua'
fail: hyde.path was not populated
fail: .../hypr/lua is not on the search path, the modules below the entry point cannot load
```

Full suite: 18 cases, 1 failed — `test_git` reports `Configs/.config/zsh/ohmyzsh is a gitlink but the tree has no .gitmodules`. That arrived with 8fcdb37d on `dev` and is untouched here.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Hyprland sessions without `HYPRLAND_CONFIG` failing to start in emergency mode.
  * HyDE now correctly locates Lua modules relative to the selected configuration file, including alternate entry points.
  * Improved reliability when loading `hyde.path` and other bundled HyDE modules.
  * Prevented configuration setups from requiring a duplicated Lua module tree.
  * Improved startup consistency across deployed and direct configuration locations.
  * Added coverage to verify module loading across supported configuration entry points.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 - Hyprland: dropped the legacy hyprlang dot, the files it deployed no longer exist
 
 ### Fixed
+- Hyprland: a session started without `HYPRLAND_CONFIG`, such as one launched from a TTY or a display manager, no longer trips emergency mode with `module 'lua.hyde.path' not found`; Hyprland resolves `require` against the directory of the config it picked, so the entry point now finds the modules shipped beside it from its own path instead of from a search path only one of the two entry points provides
 - Weather Applet: Avoid crashes from unknown weather codes or unavailable wttr.in responses.
 - Core: an install no longer ends at the Lua step on a machine that cannot build `lgi`; the introspection headers it compiles against are declared as a dependency, and an optional rock that still fails is reported and skipped instead of taking the run down before any dotfile is deployed
 - Hyprland: a session on a machine with a discrete NVIDIA GPU no longer comes up with a timed-out configuration; driver detection reads `/proc` and `/sys`, and the library directories are found by opening the candidate paths, so the budget Hyprland allows the whole configuration is no longer spent waiting on a subprocess
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 - Hyprland: dropped the legacy hyprlang dot, the files it deployed no longer exist
 
 ### Fixed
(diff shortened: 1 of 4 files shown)


kRHYME7 (changes requested):
Comments are too verbose. Let's make it concise as much as posible.

RAprogramm:
Cut down in `8270d1bf`. The block above the bootstrap went from seventeen lines to four, and the same pass went through the harness and the test:

```lua
-- require() resolves against the directory of the config Hyprland was started
-- with, which since v26.8.1 is the user's hyprland.lua, not this file. The
-- resolver is therefore loaded by its own path; everything below loads by name
-- from the search path it sets.
```

Net 33 lines of comment removed, code unchanged. Thanks for the branch update — suite is green on it, 19 cases, 0 failed, checked here against the same Ubuntu image and tool versions the workflow uses.

Issue #1930, opened 2026-08-03
[Bug]: every workflow run fails at checkout since v26.8.1 — a submodule entry with no .gitmodules


RAprogramm:
## What happens

Every workflow run in this repository has failed since 8fcdb37d, on `dev` and on every pull request, before a single check gets to run:

```
##[error]fatal: No url found for submodule path 'Configs/.config/zsh/ohmyzsh' in .gitmodules
##[error]The process '/usr/bin/git' failed with exit code 128
```

`actions/checkout` fails there, so the test suite, the shell checks and the Lua checks never start. A pull request shows red with no output to explain it.

Last run that passed: bca4bd99. First that failed: 8fcdb37d.

## Cause

`bump core hyde scripts to version v26.8.1` (8fcdb37d) meant to change one line in `Scripts/dots/hyde.toml`. It also committed a gitlink:

```
diff --git a/Configs/.config/zsh/ohmyzsh b/Configs/.config/zsh/ohmyzsh
new file mode 160000
+Subproject commit 9e23925b8581d22033f07f1983587412d3761494
```

There is no `.gitmodules` in the tree, so git has a submodule path and no url for it. `actions/checkout` walks submodules while removing its auth header, and stops there.

The path is not something HyDE tracks: `restore_shl.sh` clones oh-my-zsh into `$HOME/.oh-my-zsh` or takes it from the `oh-my-zsh-git` package, and nothing reads `Configs/.config/zsh/ohmyzsh`. It is a local clone that was picked up by `git add`. The changelog already carries `Repo: dropped two stray gitlinks that made git submodule fail on a fresh clone` — this is the same thing arriving again.

## Reproduce

A clone succeeds; anything that walks submodules afterwards does not, which is what `actions/checkout` does:

```
$ git clone https://github.com/HyDE-Project/HyDE -b dev && cd HyDE
$ git submodule status
fatal: no submodule mapping found in .gitmodules for path 'Configs/.config/zsh/ohmyzsh'
$ echo $?
128

$ git submodule foreach --recursive true
fatal: No url found for submodule path 'Configs/.config/zsh/ohmyzsh' in .gitmodules
$ echo $?
128
```

`tests/test_git.sh` reports it locally:

```
test_git
    fail: Configs/.config/zsh/ohmyzsh is a gitlink but the tree has no .gitmodules
```

## Fix

Drop the gitlink from the index. Nothing deploys from that path and nothing else in the tree refers to it.

[...]

PR #1933, opened 2026-08-03
#1931 fix: put the user's bin directory on PATH in fish

### LINKED ISSUE #1931: [Bug]: fish never gets ~/.local/bin on PATH, so hyde-shell is not found
## What happens

On fish, no HyDE command resolves:

```fish
$ hyde-shell version
fish: Unknown command: hyde-shell
```

Reported from a live session in #1897.

## Cause

`Configs/.config/fish/conf.d/hyde.fish:85`:

```fish
fish_add_path $HOME/.local/bin:$PATH
```

`fish_add_path` takes one directory per argument and does not split on `:`. The whole colon-joined string is therefore a single path named `/home/you/.local/bin:/usr/bin:/bin:…`, which does not exist, and `fish_add_path` drops paths that do not exist without a word. `~/.local/bin` never reaches `PATH`.

Checked against fish 4.8.1:

```
$ fish -c 'set -gx PATH /usr/bin /bin; fish_add_path $HOME/.local/bin:$PATH; echo $PATH'
/usr/bin /bin

$ fish -c 'set -gx PATH /usr/bin /bin; fish_add_path $HOME/.local/bin; echo $PATH'
/home/you/.local/bin /usr/bin /bin
```

Everything HyDE installs for the user lives in `~/.local/bin` — `hyde-shell`, `hydectl`, the wrappers the keybinds call — so on fish they are reachable only from a session that already had the directory on `PATH` for another reason. The zsh drop-in does it correctly in `conf.d/hyde/env.zsh`.

## Fix

```fish
fish_add_path $HOME/.local/bin
```

RAprogramm:
Closes #1931.

On fish no HyDE command resolves, which is what was reported from a live session in #1897:

```
$ hyde-shell version
fish: Unknown command: hyde-shell
```

`Configs/.config/fish/conf.d/hyde.fish:85` read

```fish
fish_add_path $HOME/.local/bin:$PATH
```

`fish_add_path` takes one directory per argument and does not split on `:`. The whole joined string is therefore a single path named `/home/you/.local/bin:/usr/bin:/bin:…`, which exists nowhere, and a path that does not exist is dropped without a word. `~/.local/bin` never reaches `PATH`, so `hyde-shell`, `hydectl` and every wrapper the keybinds call are reachable only from a session that already had the directory there for some other reason. The zsh drop-in gets this right in `conf.d/hyde/env.zsh`.

Checked against fish 4.8.1:

| Call | PATH afterwards |
| --- | --- |
| `fish_add_path $HOME/.local/bin:$PATH` | unchanged |
| `fish_add_path $HOME/.local/bin` | the directory, first |

## Tests

`tests/test_fish_env.sh` is new. It sources the shipped drop-in in an isolated home — so the universal variables `fish_add_path` writes stay out of the one running the suite — and checks that `~/.local/bin` is on `PATH` and that no entry holds a colon-joined string, which is the shape of this defect. It skips where fish is not installed.

Against this branch:

```
test_fish_env
    4 PATH entry(ies) checked
  ok
```

Against `dev`:

```
test_fish_env
    fail: the drop-in did not put ~/.local/bin on PATH, so no HyDE command resolves in fish
```

The suite here reports 19 cases with one failure, `test_git`, which is the checkout breakage #1930 covers and is untouched by this branch.

## Summary by CodeRabbit

- **Bug Fixes**
  - Restored `~/.local/bin` to the Fish shell `PATH`, allowing `hyde-shell` and keybind commands to resolve correctly.
  - Ensured PATH entries are handled separately for more reliable Fish sessions.

- **Tests**
  - Added coverage to verify Fish PATH configuration and prevent incorrectly joined entries.

- **Documentation**
  - Documented Fish PATH handling behavior in the configuration changelog.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -24,7 +24,7 @@ jobs:
       - name: Install test dependencies
         run: |
           sudo apt-get update
-          sudo apt-get install -y --no-install-recommends lua5.4 shellcheck
+          sudo apt-get install -y --no-install-recommends lua5.4 shellcheck fish
           sudo ln -sf /usr/bin/lua5.4 /usr/local/bin/lua
           sudo ln -sf /usr/bin/luac5.4 /usr/local/bin/luac
 
@@ -32,6 +32,7 @@ jobs:
         run: |
           lua -v
           shellcheck --version
+          fish --version
           python3 --version
 
       - name: Run the test suite
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 - Hyprland: dropped the legacy hyprlang dot, the files it deployed no longer exist
 
 ### Fixed
+- Fish: `~/.local/bin` reaches `PATH` again, so `hyde-shell` and everything the keybinds call resolve in a fish session; the directory was handed to `fish_add_path` joined to the rest of `PATH` by colons, which is one path that exists nowhere and is dropped without a word
 - Weather Applet: Avoid crashes from unknown weather codes or unavailable wttr.in responses.
 - Core: an install no longer ends at the Lua step on a machine that cannot build `lgi`; the introspection headers it compiles against are declared as a dependency, and an optional rock that still fails is reported and skipped instead of taking the run down before any dotfile is deployed
 - Hyprland: a session on a machine with a discrete NVIDIA GPU no longer comes up with a timed-out configuration; driver detection reads `/proc` and `/sys`, and the library directories are found by opening the candidate paths, so the budget Hyprland allows the whole configuration is no longer spent waiting on a subprocess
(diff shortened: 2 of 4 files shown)


RAprogramm:
Valid, and fixed in `d36d2140`.

The run is started with a PATH of its own, so calling the interpreter by name only worked where it happens to sit in one of the two directories that PATH holds. It is resolved before the environment is isolated now, and the resolved path is what runs — checked by putting fish somewhere else entirely and pointing PATH at that instead:

```
$ PATH=/tmp/altbin:/usr/bin:/bin sh tests/run.sh fish_env
test_fish_env
    4 PATH entry(ies) checked
  ok
```

Two things came out of looking at it again, both in the same commit.

The workflow installs `lua5.4` and `shellcheck` and no fish, so this case would have skipped in CI and the guard would have bought nothing where it matters most. `fish` joins the install step and its version is reported alongside the others.

The colon check judged every entry on PATH, including whatever the machine's own fish configuration contributes. It now looks only at the entries under the test's home directory, which are the ones this drop-in is answerable for.

PR #1944, opened 2026-08-06
fix: change switch keyboard layout target to current

time-slayer:
# Pull Request

### The Problem
Currently, `keyboardswitch.sh` uses `hyprctl switchxkblayout all next`. This creates two distinct issues, particularly for users with multiple input devices:

1. **Device Desync**. Sending `all next` does not synchronize devices to the same layout; it independently cycles every device forward one step. Two keyboards starting on different layouts will remain permanently desynced.
2. **Notification Race Condition**. When devices are in a desynced state, Hyprland takes longer to calculate and apply the next layout for all internal devices. This exposes a race condition where the subsequent  `jq` command fetches the old layout before Hyprland finishes its update, causing `notify-send` to display the incorrect language.

### The Solution
Changed the target in the switch command from `all` to `current`:
```bash
hyprctl switchxkblayout current next
```

**Why this is better**:
- Fixes the notification bug natively: modifying a single `current` device executes instantly. This eliminates the race condition without needing to artificially slow down the script with `sleep` commands.
- Eliminates phantom desyncs: it ignores secondary devices entirely, preventing them from secretly cycling layouts and causing extra processing overhead.
- Better multi-keyboard handling: if a user has an internal laptop keyboard and an external USB keyboard, triggering the shortcut will only cycle the keyboard they are currently typing on, ensuring predictable behavior.

## Type of change

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Keyboard layout switching now applies only to the active keyboard.
  * Language change notifications now appear reliably.

* **Documentation**
  * Added an unreleased changelog entry describing these corrections.


time-slayer:
Oops, I accidentally closed this for a second while trying to fix a typo in the branch name

kRHYME7 (approved):

PR #1948, opened 2026-08-07
fix: read monitor scale as a number, not by stripping the dot

### LINKED ISSUE #1946: Theme/wallpaper pickers compute columns:266 on Hyprland 0.56.2 (scale parsed as integer)
### Description

`hyde-shell theme.select` and `hyde-shell wallpaper --select` compute a nonsense rofi column count on Hyprland 0.56.2, making both pickers unusable.

The scripts derive the column count from the focused monitor's scale, parsed by stripping the decimal point out of `hyprctl -j monitors`:

```bash
mon_scale=$(jq '.[] | select(.focused==true) | .scale' <<< "$mon_data" | sed "s/\.//")
mon_scale=${mon_scale:-1}
mon_x_res=$((mon_x_res * 100 / mon_scale))
```

That assumes scale is serialised as `1.00`, so removing the dot yields `100` (a percentage). **Hyprland 0.56.2 emits an integer instead:**

```console
$ hyprctl monitors -j | jq '.[] | select(.focused==true) | .scale'
1
```

`sed` finds no dot, `mon_scale` stays `1`, and `1920 * 100 / 1 = 192000`. With `elm_width = (23+12+1) * 10 * 2 = 720`, the result is `(192000 - 40) / 720 = 266` columns instead of 2.

The `${mon_scale:-1}` fallback has the same bug — it means scale 0.01, not 1.0. It should be `100`.

### Steps to Reproduce

1. Run HyDE on Hyprland 0.56.2 with a single 1920x1080 display at scale 1
2. Confirm the scale serialisation: `hyprctl monitors -j | jq '.[].scale'` → `1`, not `1.00`
3. Run `hyde-shell theme.select`
4. Inspect the spawned process: `pgrep -a rofi`

### Expected Behavior

A usable grid. On 1920px at scale 1 the formula intends `columns:2` for the theme picker and `columns:4` for the wallpaper picker.

### Actual Behavior

```
rofi -dmenu ... -theme-str listview{columns:266;} element{...} element-icon{size:23em;} ...
```

266 columns of 23em icons. Only a few oversized thumbnails are visible and the list cannot be navigated meaningfully. Same for `hyde-shell wallpaper --select`.

### Environment

- HyDE: `8fa0073`
- Hyprland: 0.56.2
- Distro: Arch Linux
- Display: single 1920x1080, scale 1, transform 0
- rofi: 2.0.0 (wayland)

### Additional Information

Affects three scripts:

- `Configs/.local/lib/hyde/theme.select.sh`
- `Configs/.local/lib/hyde/rofiselect.sh`
- `Configs/.local/lib/hyde/wallpaper/select.sh`

Suggested fix — convert numerically rather than by string surgery, so any serialisation works (`1`, `1.00`, `1.5`, `2`):

```bash
mon_scale=$(jq '.[] | select(.focused==true) | .scale' <<< "$mon_data" \
  | awk '{printf "%d", ($1>0 ? $1*100 : 100)}')
mon_scale=${mon_scale:-100}
```

Verified locally on all three scripts: theme picker returns to `columns:2`, wallpaper picker to `columns:4`.

Happy to open a PR if the approach looks right.

Marouan-chak:
## Description

Fixes #1946.

The theme, wallpaper and rofi selectors derive their rofi column count from the focused monitor's scale, converting it to a percentage by deleting the decimal point:

```bash
mon_scale=$(jq '.[] | select(.focused==true) | .scale' <<< "$mon_data" | sed "s/\.//")
mon_x_res=$((mon_x_res * 100 / mon_scale))
```

That relies on `hyprctl -j monitors` printing `1.00`. **Hyprland 0.56.2 prints `1`** — an integer, no decimal point:

```console
$ hyprctl monitors -j | jq '.[] | select(.focused==true) | .scale'
1
```

`sed` finds no dot, `mon_scale` stays `1`, and the width is inflated a hundredfold: `1920 * 100 / 1 = 192000`. With `elm_width = (23+12+1) * 10 * 2 = 720` that asks rofi for `(192000 - 40) / 720 = 266` columns instead of 2, and the pickers render as a few oversized thumbnails that cannot be navigated.

This replaces the string surgery with a numeric conversion, so any serialisation works:

| reported scale | before | after |
|---|---|---|
| `1` | 1 ❌ | 100 ✅ |
| `1.00` | 100 ✅ | 100 ✅ |
| `1.5` | 15 ❌ | 150 ✅ |
| `2` | 2 ❌ | 200 ✅ |

Note `1.5` was already wrong before 0.56.2 — stripping the dot from `1.5` gives `15`, i.e. a scale of 0.15.

The `${mon_scale:-1}` fallback had the same bug (it means scale 0.01, not 1.0) and becomes `100`.

## Type of change

## Checklist

## Testing

Arch, Hyprland 0.56.2, single 1920x1080 at scale 1.

Before — `pgrep -a rofi` while the picker is open:

```
rofi -dmenu ... -theme-str listview{columns:266;} ... element-icon{size:23em;}
```

After:

```
theme.select      -> columns:2
wallpaper --select -> columns:4
```

Conversion checked against every serialisation in the table above, and `bash -n` passes on all three scripts.

## Notes for reviewers

- The commit also removes trailing whitespace on two lines of `rofiselect.sh`. That was `pre-commit run --all-files` acting on a file the fix already touches, not a deliberate change. Running it across the repo modifies ~200 unrelated files; those were all reverted so this PR stays focused.
- `luacheck` is not installed in my environment, so that hook was skipped. The change is shell-only.

## Summary by CodeRabbit

* **Bug Fixes**
  * Restored correct selector grid layouts on Hyprland 0.56.2.
  * Improved handling of whole-number monitor scaling.
  * Fixed display scaling and resolution normalization in selector and wallpaper views.
  * Added safe defaults when monitor scale information is unavailable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 - Hyprland: dropped the legacy hyprlang dot, the files it deployed no longer exist
 
 ### Fixed
+- Theme, wallpaper and rofi selectors show a usable grid again on Hyprland 0.56.2, which reports a whole-number monitor scale; the pickers were reading that as a hundredth of its value and laying out hundreds of columns
 - Hyprland: a session started without `HYPRLAND_CONFIG`, from a TTY or a display manager, no longer trips emergency mode with `module 'lua.hyde.path' not found`; the entry point resolves the modules shipped beside it from its own path
 - Fish: `~/.local/bin` reaches `PATH` again, so `hyde-shell` and everything the keybinds call resolve in a fish session; the directory was handed to `fish_add_path` joined to the rest of `PATH` by colons, which is one path that exists nowhere and is dropped without a word
 - Weather Applet: Avoid crashes from unknown weather codes or unavailable wttr.in responses.
--- a/Configs/.local/lib/hyde/rofiselect.sh
+++ b/Configs/.local/lib/hyde/rofiselect.sh
@@ -13,6 +13,8 @@ icon_border=$((elem_border - 5))
 mon_data=$(hyprctl -j monitors)
 mon_x_res=$(jq '.[] | select(.focused==true) | if (.transform % 2 == 0) then .width else .height end' <<< "$mon_data")
 mon_scale=$(jq '.[] | select(.focused==true) | .scale' <<< "$mon_data" | awk '{printf "%d", ($1>0 ? $1*100 : 100)}')
+mon_x_res=${mon_x_res:-1920}
+mon_scale=${mon_scale:-100}
 mon_x_res=$((mon_x_res * 100 / mon_scale))
 elm_width=$(((20 + 12 + 16) * font_scale))
 max_avail=$((mon_x_res - (4 * font_scale)))
--- a/Configs/.local/lib/hyde/rofiselect.sh
+++ b/Configs/.local/lib/hyde/rofiselect.sh
@@ -12,16 +12,16 @@ elem_border=$((hypr_border * 5))
 icon_border=$((elem_border - 5))
 mon_data=$(hyprctl -j monitors)
 mon_x_res=$(jq '.[] | select(.focused==true) | if (.transform % 2 == 0) then .width else .height end' <<< "$mon_data")
-mon_scale=$(jq '.[] | select(.focused==true) | .scale' <<< "$mon_data" | sed "s/\.//")
+mon_scale=$(jq '.[] | select(.focused==true) | .scale' <<< "$mon_data" | awk '{printf "%d", ($1>0 ? $1*100 : 100)}')
 mon_x_res=$((mon_x_res * 100 / mon_scale))
 elm_width=$(((20 + 12 + 16) * font_scale))
 max_avail=$((mon_x_res - (4 * font_scale)))
 col_count=$((max_avail / elm_width))
 [[ $col_count -gt 5 ]] && col_count=5
-r_override="window{width:100%;} 
+r_override="window{width:100%;}
     listview{columns:$col_count;}
     element{orientation:vertical;border-radius:${elem_border}px;}
-    element-icon{border-radius:${icon_border}px;size:20em;} 
+    element-icon{border-radius:${icon_border}px;size:20em;}
     element-text{enabled:false;}"
 RofiSel=$(find -L "$rofiStyleDir" -type f -exec grep -l "Attr.*launcher.*" {} \; | while
     read -r file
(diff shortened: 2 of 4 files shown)


Marouan-chak:
Good catch, thanks — that one is real and predates this PR.

`rofiselect.sh` never defaulted `mon_x_res`/`mon_scale`, unlike `theme.select.sh` and `wallpaper/select.sh` which both do. Confirmed the failure:

```console
$ mon_x_res=""; mon_scale=""; echo $((mon_x_res * 100 / mon_scale))
bash: mon_x_res * 100 / mon_scale: division by 0 (error token is "mon_scale")
```

Added the same two fallbacks the other selectors use, in 48b699b. With no focused monitor it now resolves to 1920 instead of aborting.

kRHYME7:
Will you reference the change from hyprland please? it should count as a breaking change form hyprland, imo.

kRHYME7 (approved):

Marouan-chak:
Closing this — the fix landed upstream independently while it was open, and better than what I had.

d3d30c03 added `get_monitor_scale()` to `globalcontrol.sh` and routed all three selectors through it. That is strictly better than my version: one shared helper instead of the same `awk` duplicated three times, it self-queries `hyprctl` when given nothing, and `scale * 100 + 0.5` rounds where my `%d` truncated (so 1.15 gives 115, not 114). All four files in this PR now conflict and there is nothing left worth keeping.

@kRHYME7 — the reference you asked for:

**hyprwm/Hyprland#15628** — "Don't round/truncate `hyprctl monitors` scale", by @frebib, merged 2026-07-30, shipped in **v0.56.2**.

```diff
  src/ipc/s1/Commands.cpp
-    "scale": {:.2f},
+    "scale": {},
```

`{:.2f}` always emitted two decimals, so scale 1 printed as `1.00`. Removing it prints `1`. Anything parsing the old fixed-width format sees a different string.

I agree it should have been flagged. Their reasoning is sound — `{:.2f}` was destroying precision for fractional scales, and the PR gives a concrete case: `3840 / 1.33333 = 2880` but `3840 / 1.33 = 2887.22`. So the change is correct. It just silently altered a format that tools read.

Worth noting v0.56.0's notes open with "No breaking changes! :)" and this shipped in 0.56.2 without a mention, which is exactly the gap you are pointing at.

One piece of this PR is not covered by d3d30c03: `rofiselect.sh` still has no `mon_x_res` fallback while the other two selectors default it to 1920. No longer a crash now that `get_monitor_scale` guarantees a non-zero divisor, but an empty width still evaluates to 0 and the grid collapses. Split out into #1950.

PR #1965, opened 2026-08-11
fix: restore layer blur

time-slayer:
## Description

Restores Hyprland background blur for transparent layer surfaces (Waybar, Rofi, etc.)

## Background
Previously, in [v26.7.4](https://github.com/HyDE-Project/HyDE/tree/v26.7.4), the file `Configs/.config/hypr/windowrules.conf` successfully overrode the base layer rules in `.local`, making layer blurs work as expected using this explicit configuration:
```hyprlang
layerrule = blur true,match:namespace rofi
layerrule = ignore_alpha 0,match:namespace rofi
layerrule = blur true,match:namespace notifications
layerrule = ignore_alpha 0,match:namespace notifications
layerrule = blur true,match:namespace swaync-notification-window
layerrule = ignore_alpha 0,match:namespace swaync-notification-window
layerrule = blur true,match:namespace swaync-control-center
layerrule = ignore_alpha 0,match:namespace swaync-control-center
layerrule = blur true,match:namespace logout_dialog
```
Since the migration to Lua, HyDE relies solely on `layer_rules.lua` without these overrides, causing it to use `ignore_alpha = true`. This value alters how transparency is handled and breaks the expected background blur on these layer surfaces.

## Formatting changes
Reformatted the `hl.layer_rule` calls for convenience to match [Hyprland Wiki examples](https://wiki.hypr.land/Configuring/Basics/Window-Rules/#examples).

## Type of change

## Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Restored background blur for Hyprland UI layers, including Waybar and Rofi.
  - Updated layer handling to correctly apply blur and no-animation behavior.
  - Refined transparency behavior for selection overlays and logout dialogs.

- **Documentation**
  - Added an entry documenting the Hyprland background-blur fix.


time-slayer:
The recent CI test failures appear to be caused by the latest merge from `dev` (specifically commit f143242edc7132cecef82d4b7305cbeeb996f111)

kRHYME7:
blur should be handled by the client not hyprland. So I did a trick in the past. if client layer has alpha 0 then do not blur, if it has atleast 0.1 then blur it. this way we got nice blur. 


But this might help your configs. Just create your own file and requir it inside `~/.config/hypr/hyprland.lua`

time-slayer:
Thanks for the context on how blur should ideally be handled! I completely understand the philosophy of letting the client manage it.

I just wanted to clarify the exact behavior happening right now: the current base config sets `ignore_alpha = true`. This effectively disables the compositor blur for the semi-transparent UI layers. Setting it to `0` is what allows the semi-transparent pixels to correctly trigger the background blur.

The main reason I opened the pr is that the official HyDE README (under the STYLES section) showcases background blurs on these exact elements (like `rofi`, `wlogout`). 

Here is a comparison of how it looks out of the box right now versus the expected style from the README:

### Current (`ignore_alpha = true`)
image

---

### Expected (from README)
image

---

### Fixed (`ignore_alpha = 0`)

image

---

Changing the value to `0` simply restores the visual behavior shown in those official screenshots out of the box. If the project is intentionally moving away from that default style, I totally understand and will just use a personal override in `hyprland.lua`! But if the goal is to keep the defaults matching the README, this pr fixes the current visual state.

time-slayer:
Also, even if the current design intention is to avoid applying blur by default, having semi-transparent surfaces by default without background blur makes reading text quite uncomfortable and difficult against busy wallpapers.

Additionally, from a purely technical perspective, setting `ignore_alpha = true` goes against the official Hyprland documentation. As shown in the [Hyprland Wiki Effects section](https://wiki.hypr.land/Configuring/Basics/Window-Rules/#effects-1), `ignore_alpha` expects a `number` (a float from 0 to 1), not a `boolean`.

kRHYME7:
yeah. Thanks!

PR #1968, opened 2026-08-12
refactor: generate the theme state for lua only

RAprogramm:
Split out of #1960.

Removes the theme sanitizer, `themes/theme.conf`, `themes/wallbash.conf`, the autoreload branch, the flavour check, the `hypr.config.sanitize` setting and `dots-legacy.toml`. `Scripts/migrations/v26.8.3.sh` backs up the retired files and comments out any `source =` line still pointing at them; it refuses to touch anything while the Lua entry point is unreadable, the way the migration before it does.

hyprlang stays where it is still needed: hyprlock sources `themes/colors.conf`, and themes ship `hypr.theme` in hyprlang.

## Merge order

This is the last of three stacked changes. Merge them in this order:

1. **#1960** — the fix.
2. **#1969** — the suites that cover it.
3. **#1968 — this one.** The legacy removal.

Base is `colour-state-tests` and moves to `dev` once the two below it land.

`bash tests/run.sh`: 29 cases, 0 failed.

PR #1969, opened 2026-08-13
#1960 test: cover the theme colour state and the colour pass

RAprogramm:
Split out of #1960 so that change carries the fix alone.

- `tests/test_theme_state.sh` runs the shared helpers against private homes and reads the call sites that consume them.
- `tests/test_colour_pass.sh` runs the wallpaper hand-off and the template renderer against stubs, including every failure path.
- `codecov.yml` and the coverage job in `.github/workflows/test.yml`.

## Merge order

This is the second of three stacked changes. Merge them in this order:

1. **#1960** — the fix.
2. **#1969 — this one.** These suites exercise the behaviour #1960 introduces, so merging them first would fail.
3. **#1968** — the legacy removal.

Base is `color-state` and moves to `dev` once #1960 lands.

`bash tests/run.sh`: 29 cases, 0 failed.


kRHYME7:
Let me know if ready.

RAprogramm:
Ready. Merge order: #1960 → #1969 → #1968.

Issue #1863, opened 2026-07-29
changing shaders, animations and workflows does not work


ShailajDahal-khwopa:
### Description

changing shaders, animations and workflows shows the success notification but nothing actually happens.

### Steps to Reproduce

1. Open animation, shader or workflows rofi: e.g.`hyde-shell shaders -S `
2. select an option

### Expected Behavior

Changes to a new animation, shader or workflow style

### Actual Behavior

No changes occur

### Environment

Hyprland 0.56.1 built from branch v0.56.1 at commit 5c9377c15f85c50648f35ca5a213754f95b93ca0 clean (version: bump to 0.56.1).
Date: Mon Jul 27 16:33:49 2026
Tag: v0.56.1, commits: 7643

Libraries:
Hyprgraphics: built against 0.5.1, system has 0.5.1
Hyprutils: built against 0.14.0, system has 0.14.0
Hyprcursor: built against 0.1.13, system has 0.1.13
Hyprlang: built against 0.6.8, system has 0.6.8
Aquamarine: built against 0.14.0, system has 0.14.0

Version ABI string: 5c9377c15f85c50648f35ca5a213754f95b93ca0_aq_0.14_hu_0.14_hg_0.5_hc_0.1_hlg_0.6
no flags were set

System Information:
System name: Linux
Node name: archlinux
Release: 7.1.5-arch1-2
Version: #1 SMP PREEMPT_DYNAMIC Tue, 28 Jul 2026 13:49:51 +0000

Libraries:
Hyprgraphics: built against 0.5.1, system has 0.5.1
Hyprutils: built against 0.14.0, system has 0.14.0
Hyprcursor: built against 0.1.13, system has 0.1.13
Hyprlang: built against 0.6.8, system has 0.6.8
Aquamarine: built against 0.14.0, system has 0.14.0

GPU information: 
01:00.0 VGA compatible controller [0300]: NVIDIA Corporation AD107M [GeForce RTX 4060 Max-Q / Mobile] [10de:28e0] (rev a1) (prog-if 00 [VGA controller])
06:00.0 VGA compatible controller [0300]: Advanced Micro Devices, Inc. [AMD/ATI] HawkPoint1 [1002:1900] (rev c5) (prog-if 00 [VGA controller])
NVRM version: NVIDIA UNIX Open Kernel Module for x86_64  610.43.03  Release Build  (root@)  

os-release: NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=arch
[...]

PR #1974, opened 2026-08-13
fix(workflows): apply the selected workflow immediately

kushiemoon-dev:
## Summary

Fixes the workflows part of #1863 ("changing shaders, animations and workflows does not work"): selecting a workflow from the rofi menu (`hyde-shell workflows -S`) showed the success toast but never actually changed anything.

Root cause: `rofi.workflows.lua` only wrote the selector's state file (`wf.set(selected)`). A workflow's actual effect (`hl.config(...)`, window/layer rules) lives in the workflow's own Lua file and only runs when `hyde.lua`'s `check_require("lua_state.workflows")` re-executes it with the `hl` table in scope, which happens on `hyprctl reload`, not on state write. Nothing in the selection path triggered that reload, so the new workflow only took effect on the next unrelated reload.

## Fix

Trigger `hyprctl reload` right after a successful `wf.set()`, mirroring what the `--reload` CLI flag in `luautils/selector/common.lua` already does for this exact reason.

## Test plan

- [x] `lua workflows.lua --set gaming && hyprctl reload` → `hyprctl getoption general:gaps_out` goes from `20 20 20 20` to `0 0 0 0`
- [x] `lua workflows.lua --set default && hyprctl reload` → reverts to `20 20 20 20`
- [x] Confirmed via rofi selection on a live Hyprland session that the workflow now applies on click, no manual reload needed

## Disclosure

Investigated and written with Claude Code (root-caused via reading `common.lua`/`workflows.lua`/`hyde.lua` and reproducing the missing-reload behavior with `hyprctl`), reviewed by me before opening.

Note: `apply_shader` in `shaders.lua` looks like a separate, deeper bug in the same issue: it compiles the shader but never applies it (the compiled `shader` local is unused). Left out of this PR since it needs its own fix and I haven't verified one; flagged in a comment on #1863.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved workflow setup by checking for existing workflow state before applying changes.
  * Hyprland now reloads only after a workflow is applied successfully.
  * Workflow application and reload errors are reported clearly when they occur.
  * Improved first-time workflow configuration to prevent unnecessary reloads and incomplete setup.
  * State-saving failures are now reported clearly instead of being silently ignored.
  * Improved handling of file-writing and closing errors during state updates.


kushiemoon-dev:
Retargeted onto `dev` per the release policy bot's note (master/dev have unrelated histories in this repo, so I cherry-picked the fix commit onto `dev` rather than rebasing; same single-line-scope diff, now mergeable).

kushiemoon-dev:
The failing check (`test_theme_state` inside "Configuration and script checks") is pre-existing on `dev` at f143242e (this PR's base, before any of my commits). Same failure on the [Tests workflow run there](https://github.com/HyDE-Project/HyDE/actions/runs/31690020395), unrelated to `rofi.workflows.lua`. Not something this PR introduces or should fix.

kRHYME7:
can you move the reload logic only the first time the file is created.?

kushiemoon-dev:
Yup im on it !

kushiemoon-dev:
@kRHYME7 Done. Turns out we don't even need a "first select vs later" flag, just a "does the state file already exist" check.

`hyde.lua` does `check_require("lua_state.workflows")` at config-load time, and `state.write()` writes that file via `require()` specifically so Hyprland's own hot-reload cache invalidation picks up future edits. So once the file has been created once and loaded, Hyprland already watches it, making an explicit `hyprctl reload` on every switch redundant. It's only needed the very first time the file gets created, since Hyprland has nothing to watch yet before that.

Exposed `state_file` on the selector module (`common.lua`) and check with `lfs.attributes` before `wf.set()` in `rofi.workflows.lua`, so reload fires only when the file didn't exist yet.

kRHYME7:
@rubiin @RAprogramm can you check this one out. please merge if okk

kRHYME7 (approved):

PR #1975, opened 2026-08-13
test(theme-state): align test_theme_state.sh with the lua-only refactor

kushiemoon-dev:
## Summary

`dev` has had a failing \"Tests\" workflow since f143242e (#1972, \"refactor: generate the theme state for lua only\"); `test_theme_state.sh` fails 10 assertions there and on every commit since. Noticed this on an unrelated PR (#1974) targeting `dev`.

## Root cause

#1971 (\"test: cover the theme colour state and the colour pass\") and #1972 were merged minutes apart the same day. #1972 intentionally removed `hyde_config_flavour` and every hyprlang branch it fed (`globalcontrol.sh`, `color/hypr.sh`, `theme.switch.sh`, `color.set.sh`) as part of dropping hyprlang support in favour of Lua-only config, but `test_theme_state.sh`, merged just before, still calls the now-deleted `hyde_config_flavour` directly and still expects `wallbash_state_is_complete` to treat a hyprlang colour include on its own as sufficient. Both assumptions predate the lua-only model #1972 establishes.

## Fix

- Drop every `hyde_config_flavour` probe: the function is gone by design, and the lua/hyprlang distinction it tested no longer exists.
- A hyprlang-style install with only the legacy colour include is now `incomplete` rather than `complete`, matching `wallbash_state_is_complete`'s unconditional requirement of the Lua state files (`lua_state/colors.lua`, `lua_state/ui.lua`) post-refactor.
- Drop the `color/hypr.sh` assertion requiring a `hyde_config_flavour` call: the file no longer branches on flavour at all (that's the point of the refactor); `test_hyprlang_leftovers.sh`, added in the same commit, already guards against hyprlang-only code reappearing.

## Test plan

- [x] `bash tests/test_theme_state.sh`: 0 failures (was 10) on `dev` @ f143242e
- [x] `sh tests/run.sh`: only remaining failure is `test_monitor_scale` (\"an unreadable scale reads as 150, not 100\"), reproduces identically on unmodified `dev` when run as root (root can read a file made unreadable via chmod), unrelated to this change

## Disclosure

Investigated and written with Claude Code (bisected the failure to f143242e by diffing it against its parent, traced the assertions back to the removed `hyde_config_flavour`), reviewed and the test run verified by me before opening.

## Summary by CodeRabbit

* **Tests**
  * Updated validation to focus on deployed Lua entry-point setup.
  * Confirmed that colour-only Hyprlang installations remain incomplete.
  * Verified that `wallbash_state_is_complete` is propagated to child processes.
  * Removed obsolete configuration-flavour checks.

### DIFF
--- a/tests/test_theme_state.sh
+++ b/tests/test_theme_state.sh
@@ -88,43 +88,10 @@ for created in ".local/state/hyde/lua_state" ".cache/hyde/wallbash" "run/hyde" "
         fail "loading the shared helpers left $created uncreated, so its template is skipped as a missing dependency"
 done
 
-expect "lua" "$(probe "$home_dirs" "/somewhere/hyde.lua" 'hyde_config_flavour' 2>/dev/null)" \
-    "a session on a Lua configuration"
-expect "hyprlang" "$(probe "$home_dirs" "/somewhere/hyprland.conf" 'hyde_config_flavour' 2>/dev/null)" \
-    "a session on a hyprlang configuration"
-
 home_entry="$work_dir/entry"
 mkdir -p "$home_entry/.local/share/hypr"
 probe "$home_entry" "" 'true' 2>/dev/null
 : >"$home_entry/.local/share/hypr/hyde.lua"
-expect "lua" "$(probe "$home_entry" "" 'hyde_config_flavour' 2>/dev/null)" \
-    "a deployed Lua entry point outside a session"
-
-home_config="$work_dir/config"
-mkdir -p "$home_config/.config/hypr"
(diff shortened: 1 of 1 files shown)


kRHYME7 (approved):

kRHYME7:
c/o @RAprogramm

RAprogramm:
> c/o @RAprogramm 

Yes, that's exactly what I missed.

kRHYME7:
please merge it when ready.

kushiemoon-dev:
Also make me know if anything need to be change ofc :3

Issue #1989, opened 2026-08-17
[Bug]: bundled grimblast calls slurp twice — area screenshots ask for the selection two times


ALi3naTEd0:
### Environment

- HyDE `master` @ 4c9a042
- Hyprland 0.56.1, `configProvider: lua`
- Arch Linux, `slurp` 1.5.0

### Summary

The bundled `grimblast` (`~/.local/lib/hyde/screenshot/grimblast`, shipped via `grimblast-0.1.0.tar.gz`) calls `slurp` twice inside `area()`, so every area screenshot asks you to select the region two times. The first selection is silently discarded and only the second one is used.

This affects `hyde-shell screenshot s` and `sf`, i.e. SUPER + S / SUPER + SHIFT + S with the current default binds.

### Steps to Reproduce

1. Press SUPER + S (or run `hyde-shell screenshot s`).
2. Drag a selection and release — nothing happens.
3. A second selection overlay appears; drag again — this one is captured.

### What's going on

`area()` in the bundled copy has two separate interactive `slurp` calls:

```bash
# ~/.local/lib/hyde/screenshot/grimblast, around line 241
GEOM="$(echo -n "$rects" | slurp -o "${SLURP_ARGS[@]}")"
# use `|` as separator for stableId label
choice="$(echo -n "$rects" | slurp "${SLURP_ARGS[@]}" -f "%x,%y %wx%h|%l")"
```

The upstream grimblast this is based on (https://github.com/hyprwm/contrib/blob/main/grimblast/grimblast) only has one, which both selects the region and formats the output:

```bash
choice="$(echo -n "$rects" | slurp -o "${SLURP_ARGS[@]}" -f "%x,%y %wx%h|%l")"
```

`GEOM` doesn't need the extra call either — a few lines below it is already derived from `choice`:

```bash
[[ -z "$WINDOW" ]] && GEOM="${choice%|*}"
```

### Suggested fix

Drop the extra `GEOM=` line and add `-o` to the remaining call, which makes the bundled copy match upstream:

```bash
choice="$(echo -n "$rects" | slurp -o "${SLURP_ARGS[@]}" -f "%x,%y %wx%h|%l")"
```

After that change `diff` against the upstream grimblast is clean, and area screenshots take a single selection again. Running it on two machines with no side effects so far.

### Additional Information

Since `grimblast` is bundled rather than tracked in `Configs/`, the fix has to go into the tarball — a local patch gets reverted on every `install.sh -r`.

[...]

PR #1990, opened 2026-08-17
fix: bump grimblast dot version to resync upstream fix

ALi3naTEd0:
## Description

Adds a `version` to the `[grimblast]` dot so existing installs pick up the current upstream script.

As requested in #1989. The section has no `version` of its own, so it inherits the global default and stays there across updates. `source` already tracks `hyprwm/contrib` `main` — which fixed the duplicate `slurp` call in hyprwm/contrib@3dcbce7 — but users who installed before that never re-sync, so `area()` still runs two interactive `slurp` calls and every area screenshot asks for the selection twice.

Bumping to `0.1.1` is enough to mark it as newer than the default and trigger the re-sync. Happy to use a different scheme if you'd rather tie it to an upstream date or commit.

Verified locally: after the bump the dot re-syncs and the shipped `grimblast` matches upstream (`diff` is clean), and area screenshots take a single selection again. Tested on two machines running master @ 4c9a042 with Hyprland 0.56.1.

## Type of change

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Existing installations now receive the corrected Grimblast version and properly re-synchronize the screenshot selector.

* **Configuration**
  * Grimblast now includes an explicit version declaration so updates are applied consistently.

* **Documentation**
  * Updated the Unreleased changelog with details about the screenshot selector fix and installation update behavior.


ALi3naTEd0:
Good catch by the bot — the Unreleased section already had an entry for the Grimblast selector fix, and I'd added a second one. Folded the version-resync cause into the existing line instead, so there's just the one entry now.

Also worth noting for context: since that entry is already in `dev`, the script content side is handled there. What this PR adds is only the version bump, which is what makes existing installs actually re-sync the dot rather than keeping the copy they were first installed with.

kRHYME7:
Can't think of better convention for non versioned  repo but maybe we can use our calver?

yy.mm.dd-n

the -n is use for iterations for that day just incase we need something rapid. Deez-dots will respect that as version bump too.

It is fine to omit `-n` if changes within that day is not relevant.

ALi3naTEd0:
Makes sense — switched it to `26.08.17`. Following the same convention as the rest of the repo is better than an arbitrary bump, and the `-n` escape hatch for same-day iterations is a nice touch.

Sanity-checked that deez-dots treats it as an upgrade from the default, in case anyone wonders about the jump from `0.1.0`:

```
compare_versions('0.1.0', '26.08.17')      -> -1
compare_versions('26.08.17', '26.08.17-2') -> -1
```

So both the initial bump and the optional same-day iteration resolve as newer. Left the `-n` off since there's nothing else landing for grimblast today.

rubiin (approved):

Issue #1952, opened 2026-08-07
[Bug]: sensor regex makes invalid json


obsidianjulua:
### Description

 gpuinfo.sh:195 — the sensors regex is (edge|Package id.*|another keyword). another keyword is a literal placeholder left in shipped code. That's unfinished upstream work, not a misconfiguration. It happens to match on amdgpu (edge) and Intel CPUs (Package id), so most HyDE users never see it. Your combo — AMD CPU (Tctl) with an Intel Arc dGPU (temp1 via i915) — hits neither branch, so temperature comes back empty and cascades into the invalid "percentage", and writes an error log every 5 seconds or so.

### Steps to Reproduce

1. Start HyDe

### Expected Behavior

I had claude fix the regex in a user file to parse correctly and get a real number back instead of 0 or errors 

### Actual Behavior

it reads 0 and writes and error log every 5 seconds

### Environment

[hyprland_systeminfo.txt](https://github.com/user-attachments/files/30839544/hyprland_systeminfo.txt)

Delcado19:
Partially verified against dev: the literal placeholder "another keyword" is gone from the sensors regex -- it's now edge|Package id.*|Tctl|Tdie, which covers AMD CPU (Tctl/Tdie) correctly. The actual case reported here -- an Intel Arc dGPU exposing its temperature via i915/temp1 -- is still not matched, so this isn't fully resolved. Given the ongoing gpuinfo.lua rewrite in #2062, the Intel Arc dGPU sensor gap probably belongs there rather than as a separate patch to the (soon to be replaced) sensors regex in gpuinfo.sh. Leaving open for that remaining part.

PR #2022, opened 2026-08-30
fix: keep gpuinfo output valid JSON on a cold start

### LINKED ISSUE #2021: [Bug]: gpuinfo.sh division by zero + stdout/stderr mixing crashes waybar JSON parsing on cold start
## Description

`Configs/.local/lib/hyde/gpuinfo.sh` has two related bugs that both cause it to spam invalid output/errors every ~5 seconds (the waybar module poll interval), which can crash the JSON parsing for `custom/gpuinfo` in waybar:

**1. Division by zero in `get_utilization()` (line ~223) on cold start**

```sh
awk -v stat="$diffStat" -v idle="$diffIdle" 'BEGIN {printf "%.1f", (stat/(stat+idle))*100}'
```

The state file (`/tmp/hyde-$UID-gpuinfo`) is cleared on every reboot (it lives in `/tmp`). On the
very first invocation after that, `GPUINFO_PREV_STAT`/`GPUINFO_PREV_IDLE` are seeded from the
*current* `/proc/stat` reading, so `diffStat` and `diffIdle` both come out `0` on that same call →
`stat/(stat+idle)` is `0/0` → `awk: fatal: division by zero attempted`, and the module falls back
to stale/empty output for that cycle.

**2. The "Initialized Variable" banner is written to stdout, not stderr (lines ~269, ~284)**

```sh
echo -e "Initialized Variable:\n$(cat "$gpuinfo_file")\n\nReboot or '$0 --reset' to RESET Variables"
```

This runs on the same cold-start path as bug 1, and its output goes to stdout — which is exactly
what waybar reads as the module's JSON. The banner text isn't JSON, so waybar's `return-type: json`
parser fails immediately with `Syntax error: value, object or array expected. Line 1, Column 1`.

Both only show up on the first call after a reboot / after `--reset` / after `/tmp` gets cleared,
which is why they're easy to miss in normal use but show up reliably right after boot, and (per
the journal) apparently also after some other event that clears/misses the state file mid-session
— haven't fully isolated that second trigger yet.

This looks like the same broad "gpuinfo.sh silently produces bad output and error-logs every 5s"
symptom family as #1952, but a different concrete cause (that one is the temperature regex on
line 195 having a literal unfinished `another keyword` placeholder for uncommon sensor layouts;
this is the CPU-utilization branch failing specifically on cold start, independent of sensor
layout).

## Steps to Reproduce

1. `rm -f /tmp/hyde-$UID-gpuinfo*` (simulates a fresh boot / cleared `/tmp`)
2. Run the module's exec command directly, e.g. `hyde-shell gpuinfo --use intel` (or whichever
   vendor `custom-gpuinfo#<vendor>.jsonc` uses on your machine)
3. Observe the first invocation prints the "Initialized Variable..." banner to stdout followed by
   `awk: fatal: division by zero attempted` on stderr, and the JSON line for that first call is
   either missing or waybar logs a JSON parse error for it

## Expected Behavior

The module should always emit exactly one valid JSON object on stdout per invocation, with `0`
(or omitted) utilization on the very first cold-start sample instead of crashing awk, and any
human-readable diagnostic text should go to stderr so it never lands in waybar's parsed stdout.

## Actual Behavior

- `awk: fatal: division by zero attempted` on the first invocation after `/tmp` state is missing
- The "Initialized Variable... Reboot or --reset" banner text is mixed into stdout, which waybar
  then fails to parse as JSON (`Syntax error: value, object or array expected`)

## Environment

<details>

```
Date: Wed Aug 5 14:13:21 2026
Tag: v0.56.2, commits: 7661

Libraries:

no flags were set


System Information:
System name: Linux
Node name: atlas
Release: 7.1.9-arch1-2
Version: #1 SMP PREEMPT_DYNAMIC Fri, 21 Aug 2026 22:18:59 +0000

Libraries:



GPU information: 
0000:00:02.0 VGA compatible controller [0300]: Intel Corporation TigerLake-LP GT2 [Iris Xe Graphics] [8086:9a49] (rev 01) (prog-if 00 [VGA controller])


os-release: NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=arch
BUILD_ID=rolling
ANSI_COLOR="38;2;23;147;209"
LOGO=archlinux-logo


plugins:

Explicit sync: supported
GL ver: 3.2
Backend: drm

Monitor info:
	Panel eDP-1: 1366x768, eDP-1 Chimei Innolux Corporation 0x1537  -> backend drm
		explicit ✔️
		edid:
			hdr ❌
			chroma ✔️
			bt2020 ❌
		vrr capable ❌
		non-desktop ❌
		

State:

configProvider: lua
backend: drm




======Config-Start======
Not supported under lua
======Config-End========
```

</details>

- Installation Environment: On the main hardware
- Display Resolution: 1366x768
- Graphics Driver: intel (also present, differently, on an nvidia machine — the same
  `general_query`/`get_utilization` code path is shared across the intel/nvidia "generic" branch)
- Display Server: Wayland

## Additional Information

Related: #1952 (same file, same "invalid JSON / error every ~5s" symptom, different root cause).

I have a minimal fix for both (guard the division, redirect the banner to stderr) and I'm happy
to open a PR against `dev` once this is triaged, unless someone already has a fix for #1952 in
flight that touches the same function.

cativo23:
# Pull Request

## Description

Fixes #2021.

On a cold start — the state file in `/tmp` cleared by a reboot, or after `--reset` — three
separate defects each break the `custom/gpuinfo` waybar module, which reads the script's stdout
as `return-type: json`:

1. **Division by zero.** `general_query` → `get_utilization` reads `/proc/stat` once, seeds
   `GPUINFO_PREV_STAT`/`GPUINFO_PREV_IDLE` from that same reading, then diffs against it — so
   both diffs are `0` on that first call and `awk` exits with `division by zero attempted`.
2. **Banner on stdout.** The same cold-start path echoes `Initialized Variable...` to stdout
   rather than stderr, so that plain text lands in what waybar parses as JSON.
3. **Empty `percentage`.** When no temperature is available, `temp_pct` is unset and the module
   emits a literal `"percentage":,` — not valid JSON.

Fix: guard the division, send both banners to stderr, and default the percentage to `0`.

Adds `tests/test_gpuinfo.sh` covering all three. It runs the script cold with no arguments and
with `sensors` shadowed by a stub on `PATH`, so `general_query()` and the missing-temperature
path both run identically regardless of the host's GPU or sensors. Verified red→green against
each defect individually.

**Note on #1952:** the third fix is a guard, not a cure for that issue. #1952 is about the
temperature regex on line 195 still carrying a literal `another keyword` placeholder, which is
why temperature comes back empty on some hardware — that root cause is untouched here. This
change only guarantees the module keeps emitting parseable JSON whatever the reason a value is
missing.

## Type of change

## Checklist

- [x] All new and existing tests passed (`sh tests/run.sh`: 30 cases, 0 failed).

## Additional context

The third defect was caught by the new test on CI, where the runner has no hardware sensors —
which is exactly the scenario it now reproduces deliberately via the `sensors` stub, so it no
longer depends on what the runner happens to have.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed the Waybar GPU information module crashing during its first poll after reboot or reset.
  * Prevented initialization messages from appearing in JSON output and breaking Waybar parsing.
  * Improved handling when utilization data is unavailable or has a zero denominator.
  * Ensured valid output when no temperature sensor is available.

* **Tests**
  * Added coverage for cold-start behavior, missing temperature sensors, and valid JSON output.


cativo45:
Thanks — both applied in the latest push.

- The stderr capture now goes through `mktemp` and is cleaned up in the existing exit trap, instead of a predictable `/tmp` name.
- The JSON assertion no longer looks at only the last line. One correction to the suggestion though: it checks that *every* non-empty line of stdout is a JSON object, rather than requiring stdout as a whole to parse as one. A cold-start run legitimately emits more than one object, because the startup detection re-invokes the script, and waybar consumes the stream line by line — so requiring a single object would fail on correct output. Checking per line still catches what the suggestion was after: any diagnostic text printed before the JSON now fails the test.

Verified red→green against each defect individually after the change: dropping either fix makes the test fail, and the missing-banner case is now caught by the JSON check as well as the text check.

Issue #1983, opened 2026-08-15
[Bug]: A few bugs in Lua update and how to fix them


a78738825:
## List of Bugs (With solutions if any. Thank me later 😉):
1. Dolphin file manager defaults to a floating window
2. Rofi launcher not getting Blur effect
3. zsh cursor behaviour (pressing arrow keys results in jumping whole words instead of single characters)
4. Animations broken
5. [Read this from @miyuki-ts](https://github.com/HyDE-Project/HyDE/issues/1983#issuecomment-5334135946)

## Solutions
1. Remove/Comment out, line `29` -> `"org\\.kde\\.dolphin",` in file: `~/.local/share/hypr/lua/window_rules.lua`.
2. Remove/Comment out, line `18` -> `"rofi",` in file `~/.local/share/hypr/lua/layer_rules.lua`.
3. Comment out everything inside `~/.config/zsh/conf.d/binds.zsh`.
4. use command `hyde-shell animations --select` and choose a different animation from the list.

> **The edits in `~/.local/share/hypr/` are temporary and will be reset after updating HyDE. Permanent changes require editing the `~/.config/hypr/hyprland.lua`**

### Steps to Reproduce

1. Use `uninstall.sh` in pre Lua HyDE
2. Use `install.sh` and update to latest version (commit `fa96c4f9d6eb33944eb6d03de9290c59208c1a7f`)

### Expected Behavior

No difference in UI, from pre Lua update HyDE.

### Actual Behavior

Broken UI with a few misc bugs

### Environment

.

### Additional Information

<img width="267" height="134" alt="Image" src="https://github.com/user-attachments/assets/504da768-7a4d-4119-991f-75e724b87e42" />

Keep forking around 😇😇

rubiin:
1 is fixed on dev

a78738825:
@rubiin any idea about how to fix 3rd? firefox not only seems to have wrong scaling, but no theming at all.
take a look:

<img width="506" height="572" alt="Image" src="https://github.com/user-attachments/assets/8c35996d-df70-4244-b7e0-c80bc95ddab9" />

<img width="733" height="768" alt="Image" src="https://github.com/user-attachments/assets/6dfdf2cb-7ccf-4a52-8f9e-cfd131396f9c" />

or is it only happening with me>

ALi3naTEd0:
Thanks for this list — the dolphin one saved me.

[...]

Issue #1953, opened 2026-08-08
[Bug]: Incorrect command reference in custom-theme.jsonc Waybar module ("hyde-shell themeselect" instead of "theme.select")


bukananda:
### Description

In the Waybar configuration for the custom-theme module (waybar/modules/custom-theme.jsonc), the on-click-middle command is set to call hyde-shell themeselect. However, in the HyDE library scripts (lib/hyde), the script responsible for opening the theme selector menu is named theme.select (or exposed via hyde-shell theme select / hyde-shell theme.select).

Because of this mismatch, clicking middle-click on the theme module in Waybar fails with a command not found or unhandled option error in hyde-shell.

### Steps to Reproduce

 hyde-shell themeselect

### Expected Behavior

Change themes

### Actual Behavior

hyde-shell themeselect                                               23:58 
Command not found: themeselect
Available commands:

  Available Scripts
  ───────────────────────────────────────────────────────────────────────────────────────────────────────
    amdgpu
    app
    battery
    bookmarks
    brightnesscontrol
    calculator
    cava
    cliphist
      cliphist.image
    color
    completions
    cpuinfo
    emoji-picker
    fastfetch
    font
    fzf_preview
    fzf_wrapper
    gamelauncher
    gamemode
    globalcontrol
    glyph-picker
    gpuinfo
    help
    hyde-launch
    hyprlock
    hyprsunset
    idle-inhibitor
    keybinds_hint
    keyboardswitch
    lockscreen
    logoutlaunch
    mediaplayer
    notifications
    pm
    polkitkdeauth
    quickapps
    reload
    resetxdgportal
    restore
    rofi
      rofi.bookmarks
      rofi.websearch
    rofilaunch
    rofiselect
    screenrecord
    screenshot
    sensorsinfo
    session
    system
      system.monitor
      system.update
    theme
      theme.import
      theme.patch
      theme.select
      theme.switch
    volumecontrol
    wallbash
      wallbash.print.colors
    wallbashtoggle
    wallpaper
      wallpaper.awww
      wallpaper.hyprpaper
      wallpaper.kon
      wallpaper.mpvpaper
      wallpaper.swww
      wallpaper.waydeeper
    waybar
    weather
    window

  ───────────────────────────────────────────────────────────────────────────────────────────────────────
  67 scripts across 1 directory

### Environment

[...]

Issue #1945, opened 2026-08-07
HyDE themes and configs broken after Hyprland Lua update (v0.56)


Leooo7262627:
### Description

After updating Hyprland to version 0.56 (Lua-based configuration), my HyDE setup became heavily broken. Multiple core features stopped working, especially themes, wallpapers, and UI components.

This appears to be caused by the migration from .conf to .lua, where HyDE and its themes are still relying on the old configuration format, leading to conflicts with the new system.

### Steps to Reproduce

1. stall HyDE on Arch Linux
2. Update the system to the latest Hyprland (v0.55+ with Lua config)
3. Apply any theme or wallpaper using HyDE
4. Run hyde-shell reload or restart the session
5. Observe broken behavior

### Expected Behavior

- Themes should apply correctly
- Wallpapers should display normally
- HyDE features (UI, lock screen, gestures, etc.) should work properly
- Configuration changes should be applied without errors

### Actual Behavior

- Themes appear blank or fail to apply
- Wallpapers do not display
- Waybar styles and fonts break
- hyde-shell reload has no effect
- Some configuration changes are ignored
- Gestures are broken (e.g., 3-finger swipe not working, fallback to 4-finger or no response)
- Lua-related errors may appear
- General instability after the update

### Environment

Installation Environment: On The Main Hardware
Display Resolution: 1920x1080
Graphics Driver: Intel (or your GPU)
Display Server: Wayland
<details> <summary>System Information</summary>

Paste output of:

hyprctl systeminfo -c
</details>

### Additional Information

- OS: Arch Linux
- Hyprland version: v0.56 (Lua-based)
- Using HyDE Project
- Issue started immediately after updating Hyprland
- Likely caused by conflicts between old .conf configs and new .lua system
- Similar issues reported by others after the Lua migration

SumitKumar-17:
Another addition to this: 
 I have been a user of Hyde for the last 1.5 years. I recently pulled and ./install.sh a few days back(since 4-5 months I believe or more maybe) . After the update the lock screen wallpaper is black, showing "[font] Layout file not found". How can I fix this.

So for the dropdown terminal "Super Alt +T " not works, it shows an error notifications "Unkown command toggle_console. Try help for available commands"


Confirm: 3-finger did not also work for me, I manually changed the code of the files to make it work as I am now habituated to it. Should I make a PR of 3-finger swipe to the repo? Or you guys have already planned it.

Leooo7262627:
[...]

PR #2024, opened 2026-08-30
fix: run theme.switch/theme.select/wallpaper.sh in their own scope, out of waybar's cgroup

### LINKED ISSUE #2025: [Bug]: Theme picker applies the theme name but not the wallpaper/colours (killed by waybar's own reload signal)
### Description

Picking a theme from the Waybar right-click menu (HyDE Options → Theme → Select Theme), or from the theme module's next/previous actions, intermittently updates the theme name (`$HYDE_THEME`) but leaves the wallpaper and colour scheme on the previous theme. No error is shown anywhere — the rofi picker closes normally and a notification for the new theme name still fires.

Root cause: applying a theme rewrites `~/.config/hypr/themes/colors.conf`. Hyprland's autoreload (`misc:disable_autoreload` defaults to `false`) picks up that change and reloads. HyDE has a hook that reacts to a Hyprland reload by running `systemctl --user kill -s SIGUSR2 hyde-Hyprland-bar.service` to refresh Waybar (`Configs/.local/lib/hyde/waybar.py`, `_signal_waybar`). `systemctl kill` defaults to `--kill-who=all`, which signals **every process in that unit's cgroup**, not just waybar's main PID. `theme.select.sh`/`theme.switch.sh` run as descendants of waybar (menu actions are spawned as its children) and inherit its cgroup, so the in-flight colour/wallpaper pipeline (`color.set.sh`, GNU `parallel`) gets `SIGUSR2`'d and dies mid-run along with waybar's own reload. The pipeline reliably takes a few seconds, so the race is easy to lose.

I opened a PR with the fix and a full writeup of how I traced it: #2024

### Steps to Reproduce

**A. Natural repro (timing-dependent, may take a couple of tries):**

1. Right-click the HyDE Options module in Waybar → Theme → Select Theme.
2. Pick a theme that's visibly different from the current one (different wallpaper).
3. Watch the desktop background.

Expected: wallpaper changes to the new theme's wallpaper.
Actual: sometimes the rofi picker closes, a notification for the new theme name appears, `hyprctl -j getoption misc:disable_autoreload` and `$HYDE_THEME`/the state file (`~/.local/state/hyde/staterc`) reflect the new theme, but the wallpaper (checkable with `swww query` / `awww query` depending on backend) never changes to match, and colours are left on the old theme too.

**B. Deterministic repro (what I used to confirm the mechanism and verify the fix):**

```bash
# run a theme switch, and fire the exact signal HyDE's own reload hook sends, mid-run
/home/<user>/.local/lib/hyde/theme.switch.sh -s "<some other theme name>" &
sleep 0.4
systemctl --user kill -s SIGUSR2 hyde-Hyprland-bar.service
wait
```

Expected: theme, colours, and wallpaper all end up matching the target theme.
Actual (before the fix in #2024): `$HYDE_THEME` and the theme's colour-state files get updated, but `awww query`/`swww query` still reports the *previous* theme's wallpaper — the pipeline died partway through `wallpaper.sh` → `color.set.sh` (visible as e.g. `color.set.sh: Zeile 1: <pid> Benutzerdefiniertes Signal 2   parallel fn_wallbash ...` in `journalctl --user`).

### Expected Behavior

Selecting a theme always applies the theme name, colours, and wallpaper together, regardless of what Waybar/Hyprland are doing at the same moment.

### Actual Behavior

The theme name can get persisted while the colour/wallpaper application is killed mid-flight by HyDE's own Waybar-reload signal, leaving the desktop in a mixed state (new theme name, old wallpaper/colours) with no error surfaced anywhere.

### Environment

- Installation Environment: On the main hardware
- Display Resolution: 1920x1080
- Graphics Driver: intel
- Display Server: Wayland

<details>

```

Libraries:

System Information:
System name: Linux
Node name: cachyos-x8664
Release: 7.2.2-1-cachyos

GPU information:
00:02.0 VGA compatible controller [0300]: Intel Corporation Kaby Lake-U GT2 [HD Graphics 620] [8086:5916] (rev 02)

os-release: NAME="CachyOS Linux", ID_LIKE=arch

State:
configProvider: lua
backend: drm
```

</details>

### Additional Information

- `theme.switch.sh` already tries to guard against a Hyprland reload happening mid-switch: it runs `hyprctl eval 'hl.config({misc = {disable_autoreload = true}})'` near the top, before writing any theme state, and never sets it back to `false`. Despite that, `hyprctl -j getoption misc:disable_autoreload` reads back `false` both at rest and while a switch is in flight on this system, and the reload still fires. I didn't chase why — this system is on Hyprland's Lua config provider (`configProvider: lua`), and there are other open reports about the Lua migration's reload/state behaviour (#1945, #1983) this may be entangled with.
- Not the same bug as #1953 (already fixed) — that one made the menu action fail to do anything at all (stale `themeselect` command name, no dialog ever opened). This bug only shows up after the picker successfully opens and a theme is chosen.
- Wallpaper backend in use: `awww`. I'd expect the same race to apply to the other backends (`swww`, `hyprpaper`, etc.) since the kill happens before the backend is even reached, but only tested with `awww`.

Delcado19:
## Description

Selecting a theme, or just a wallpaper within the current theme, from any of Waybar's theme/wallpaper menu actions updates `$HYDE_THEME`/`wall.set`, but the wallpaper and colours silently stay unapplied, roughly half the time, with no error shown anywhere.

### Root cause

1. `theme.select.sh`/`theme.switch.sh`/`wallpaper.sh` are spawned as direct children of `waybar` by its menu actions, so they inherit waybar's systemd cgroup (`hyde-Hyprland-bar.service`).
2. Applying a theme, or setting a wallpaper "as global", writes `~/.config/hypr/themes/colors.conf`.
3. Hyprland's autoreload (`misc:disable_autoreload` defaults to `false`) picks up that file change and reloads.
4. HyDE's reload hook then runs `systemctl --user kill -s SIGUSR2 hyde-Hyprland-bar.service` (`waybar.py:_signal_waybar`) to refresh waybar.
5. `systemctl kill` defaults to `--kill-who=all`, signalling **every process in the unit's cgroup**, not just waybar's main PID. The still-running script/`color.set.sh`/`parallel` tree is in that same cgroup and gets `SIGUSR2`'d along with waybar.
6. The colour pipeline (wallbash + GNU `parallel`) reliably takes a few seconds, so the reload lands mid-flight often enough to notice.

Initially I only found and fixed this for the theme picker; a follow-up report against the same root cause (plain wallpaper switching within a theme, and the standalone wallpaper widget) showed the exact same failure, since `wallpaper.sh --global` writes the same `colors.conf` via the same code path.

### Fix

Per @kRHYME7's suggestion, this launches the affected scripts through the existing `hyde-shell app -t scope --` wrapper from every waybar menu action that triggers them (HyDE menu, theme module, wallpaper widget, macOS layout menu) — matching the pattern already used for `custom-gpuinfo#nvidia.jsonc`. `theme.select.sh`'s own direct call into `theme.switch.sh` inherits the same scope, so only the outermost entry point needs wrapping in each case, and **none of the scripts themselves need changes**.

Verified with the deterministic repro from #2025 (real `systemctl --user kill -s SIGUSR2 hyde-Hyprland-bar.service` fired mid-run), for both the theme switch and the plain wallpaper switch paths:

```bash
hyde-shell app -t scope -- theme.switch.sh -s "Dracula" &      # or: wallpaper.sh -s <path> --global &
sleep 0.4
systemctl --user kill -s SIGUSR2 hyde-Hyprland-bar.service
```

Without the wrapper both reproduce the bug every time; with it, theme/wallpaper/colours all end up consistent.

### Type of change

### Checklist

## Additional context

- `theme.switch.sh` already tries to guard against exactly this by running `hyprctl eval 'hl.config({misc = {disable_autoreload = true}})'` near the top, before it writes any theme state, and never resets it back to `false` afterwards. Despite that, `misc:disable_autoreload` reads back `false` on my system both at rest and while a switch is in flight, and the reload still fires. Didn't chase this further — this system is on Hyprland's Lua config provider (`configProvider: lua`), and there are other open reports about the Lua migration's reload/state behaviour (#1945, #1983) this may be entangled with. This PR fixes the user-visible symptom regardless of whether that gets sorted out separately.
- Different bug from #1953 (already fixed) — that one made the menu action fail to do anything at all (stale `themeselect` command name, no dialog ever opened). This one only shows up after the picker successfully opens and a theme/wallpaper is chosen.
- Didn't touch `animations`/`layouts`/`workflows`/`shaders` menu actions — I don't have evidence they write files Hyprland's autoreload watches, so wrapping them would be speculative. Worth checking if similar reports come in for those.

Fixes #2025

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where switching themes could stop before wallpaper and color updates completed.
  * Improved reliability of theme and wallpaper actions triggered from Waybar menus and controls.
  * Theme selection, next-theme, and previous-theme actions now complete consistently.
  * Waybar now refreshes correctly after wallpaper or theme changes.
  * Prevented repeated `gpuinfo` error messages when battery power data is unavailable.


kRHYME7:
@Delcado19 can you check  the effects of `-t service` and `-t scope`?

Delcado19:
Tested both on my system. TL;DR: **scope is the right choice for this fix**, service would reintroduce a race.

**Cgroup isolation — identical for both:**
```
waybar unit:  .../app-graphical.slice/hyde-Hyprland-bar.service
-t scope  ->  .../app-graphical.slice/app-Hyprland-sh-e28864c8.scope
-t service -> .../app-graphical.slice/app-Hyprland-sh@bac16214.service
```
Both land in their own cgroup, separate from waybar's — so both are equally immune to `systemctl --user kill -s SIGUSR2 hyde-Hyprland-bar.service`.

**Execution contract — this is where they differ:**
```
$ time hyde-shell app -t scope -- sh -c 'echo start; sleep 2; echo end'
start
end
... 2.03s total

$ time hyde-shell app -t service -- sh -c 'echo start; sleep 2; echo end'
... 0.08s total   (no "start"/"end" printed to caller — went to journal instead)
```
`-t scope` runs `systemd-run --scope` in the foreground: it blocks the caller until the command finishes, and stdout/stderr/exit-code are inherited normally. `-t service` starts a detached transient unit and returns almost immediately, before the command has actually run — output goes to the journal, not back to the caller.

That matters here because several of the wrapped call sites are `;`-chained on the waybar side, e.g.:
```
"on-click" : "hyde-shell app -t scope -- wallpaper.sh --select --global;pkill -RTMIN+19 waybar"
```
This relies on `wallpaper.sh` (and its `colors.conf` write) actually finishing before the trailing `pkill -RTMIN+19 waybar` fires, so waybar's own refresh sees the final state. With `-t service` that ordering guarantee is gone — `pkill` would very likely fire while `wallpaper.sh`/`theme.select.sh` are still mid-run, so waybar could refresh on stale/partial state. That's the same class of problem this PR fixes, just via a race instead of a hard kill. It also matches @coderabbitai's earlier concern about preserving the synchronous completion contract.

So I kept `-t scope` in the current version of the PR — confirmed it still fully resolves the original repro (theme + wallpaper-only cases) without introducing this new race.

kRHYME7:
ah right! Actually app2unit/uwsm uses scope by default. And HyDE Added an env to use service by default (someone PR'd that and the reason is KDE does that too.) Which is better for cleaning up.

Explicitly running it as scope is better. will merge  now. Thanks! Please open a separate pr if you noticed other race conditions

PR #2027, opened 2026-08-31
fix: stop gpuinfo failing on an unreadable power_now

### LINKED ISSUE #2026: [Bug]: gpuinfo floods stderr with an awk fatal error when a battery's power_now cannot be read
## Description

`general_query()` in `Configs/.local/lib/hyde/gpuinfo.sh` guards a battery's `power_now` with `[[ -f $file ]]` and then hands the *path* to `awk`, letting awk open the file:

```sh
for file in /sys/class/power_supply/BAT*/power_now; do
    [[ -f $file ]] && power_discharge=$(awk '{print $1*10^-6 ""}' "$file") && break
done
```

The file existing doesn't mean it can be read. The ACPI battery driver exposes `power_now` unconditionally through the generic power-supply attribute group, but `acpi_battery_get_property()` returns `ENXIO` when the firmware has no such reading, and awk dies with a fatal error on every invocation.

With `custom/gpuinfo` at `"interval": 5`, that's one fatal error every 5 seconds, forever: around 17,000 lines a day in the journal, plus a wasted `awk` process per poll.

The `current_now` fallback doesn't rescue these machines either — an energy-reporting battery (`energy_now` / `energy_full`) has no `current_now` file at all.

This has been documented kernel behavior for over a decade: see [Launchpad #889942](https://bugs.launchpad.net/ubuntu/+source/linux/+bug/889942) and [ACPI Bugzilla #83411](https://acpi-bugzilla.narkive.com/uldExK7u/bug-83411-new-sys-class-power-supply-bat0-current-now-no-such-device).

**Scope:** `general_query()` is reached both from `intel_GPU()` and from the final `else` branch, so it affects any machine where no GPU vendor is detected, not just Intel laptops.

This predates #2022 and is unrelated to it — the same code is in `master` and `dev`.

## Steps to Reproduce

1. Use a machine whose battery exposes `power_now` but cannot read it — confirm with:

   ```console
   $ cat /sys/class/power_supply/BAT0/power_now
   cat: /sys/class/power_supply/BAT0/power_now: No such device
   ```

   (`python3 -c "open('/sys/class/power_supply/BAT0/power_now').read()"` reports `OSError(19, 'No such device')`, i.e. ENXIO.)

2. Run the module and look at stderr only:

   ```console
   $ hyde-shell gpuinfo 2>&1 >/dev/null
   awk: fatal: error reading input file `/sys/class/power_supply/BAT0/power_now': No such device
   ```

3. Or just let waybar run and watch the journal:

   ```console
   $ journalctl --user -u "hyde-${XDG_SESSION_DESKTOP}-bar.service" --since '24 hours ago' | grep -c 'awk: fatal'
   ```

## Expected Behavior

An attribute that can't be read should just get skipped. The tooltip omits power draw, same as what already happens when the file is absent entirely — that's the correct outcome on hardware that can't report it.

## Actual Behavior

`awk` aborts with a fatal error on stderr once per poll, forever:

```
awk: fatal: error reading input file `/sys/class/power_supply/BAT0/power_now': No such device
```

Either way, `power_discharge` ends up empty, so the visible output is already correct. The only real damage is the error spam on every single poll.

## Environment

- Graphics Driver: intel (TigerLake-LP GT2, i915)
- Display Server: Wayland
- HyDE: present on both `master` (`7c7b832d`) and `dev` (`a1eb2c05`)
- Battery: energy-based (`energy_now` / `energy_full`), **no** `current_now`

<details>

```
Date: Wed Aug 5 14:13:21 2026
Tag: v0.56.2, commits: 7661

Libraries:

no flags were set


System Information:
System name: Linux
Node name: atlas
Release: 7.1.9-arch1-2
Version: #1 SMP PREEMPT_DYNAMIC Fri, 21 Aug 2026 22:18:59 +0000

Libraries:



GPU information: 
0000:00:02.0 VGA compatible controller [0300]: Intel Corporation TigerLake-LP GT2 [Iris Xe Graphics] [8086:9a49] (rev 01) (prog-if 00 [VGA controller])


os-release: NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=arch
BUILD_ID=rolling
ANSI_COLOR="38;2;23;147;209"
LOGO=archlinux-logo


plugins:

Explicit sync: supported
GL ver: 3.2
Backend: drm

Monitor info:
	Panel eDP-1: 1366x768, eDP-1 Chimei Innolux Corporation 0x1537  -> backend drm
		explicit ✔️
		edid:
			hdr ❌
			chroma ✔️
			bt2020 ❌
		vrr capable ❌
		non-desktop ❌
		

State:

configProvider: lua
backend: drm




======Config-Start======
Not supported under lua
======Config-End========

```

</details>

## Additional Information

Separate pre-existing nit I noticed while investigating, not part of this report: `var=$(cmd)` returns the command's exit status, so when awk fails the `&&` chain stops before `break` runs. Harmless today, since the `for` loop just moves on to the next glob match, but probably worth a one-line follow-up of its own.

I have a fix and a regression test ready and will open a PR against `dev` referencing this issue.

cativo23:
## Description

`general_query()` guarded a battery's `power_now` with `[[ -f $file ]]` and then handed the path to `awk`, letting awk open the file. The file existing doesn't mean it can be read: the ACPI battery driver exposes `power_now` unconditionally but returns `ENXIO` when the firmware can't supply that value, so awk dies with a fatal error on stderr for every single invocation, once per waybar interval, forever. The `current_now` fallback doesn't help those machines either, since an energy-reporting battery has no `current_now` to read.

`general_query()` is reached from `intel_GPU()` and from the final `else` branch, so this affects any machine where no GPU vendor is detected, not just Intel laptops.

The fix reads the value first and only uses it when something came back, so awk receives a value rather than a path and has no file to fail on. The same guard is applied to the `current_now`/`voltage_now` fallback. Behavior is unchanged wherever `power_now` is readable, verified with a real value (`5432100` -> `5.4321` before and after).

The regression test needs to inject an unreadable attribute, which the hardcoded absolute glob made impossible, so the directory is overridable via `GPUINFO_POWER_SUPPLY_DIR`. It defaults to `/sys/class/power_supply` and is only ever read, so shipped behavior is identical.

Fixes #2026

**On the test, honestly:** the mode-000 injection point only exists because of this fix. Running that case against the unfixed script doesn't exercise the mock — it's a guard against reintroducing the defect, not a from-scratch reproduction of the original bug. A mode-000 file also fails with `EACCES` rather than the `ENXIO` a real unsupported `power_now` returns, but both reach the same failure (awk dies, value comes back empty), which is what the guard actually checks. That caveat is written into the test's comment block too.

**Not addressed here:** because `var=$(cmd)` returns the command's exit status, when awk fails the `&&` chain stops before `break` runs. Pre-existing, harmless today, and left alone to keep this PR to one concern.

## Type of change

## Checklist

## Additional context

Full suite passes locally (30 cases, 0 failed). On the affected hardware the `awk: fatal` rate went from one every 5 seconds to zero. The same buggy code is present in `master` (`7c7b832d`), so this will need to reach there via the usual release path.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed the Waybar GPU information module flooding stderr with `awk` errors when battery power data is unavailable or unreadable.
  * Improved handling of missing or inaccessible battery readings so GPU information continues displaying without fatal errors.
  * Prevented repeated polling errors when firmware cannot provide battery power data.

* **Tests**
  * Added coverage for unreadable battery power attributes to help prevent regressions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 - Hyprland: dropped the legacy hyprlang dot, the files it deployed no longer exist
 
 ### Fixed
+- Waybar: `gpuinfo` no longer floods stderr with an `awk` fatal error on every poll when a battery exposes a `power_now` attribute the firmware cannot actually read; the value is now read before it is used instead of being handed straight to `awk`
 - Installer: a fresh install no longer aborts on a missing AUR helper before having the chance to install it
 - Wallbash: resolve `integer expected` syntax error in `color.set.sh` when evaluating template failure state
 - Waybar: resolved an issue in the memory module where state-specific formats overrode `format-alt` when memory usage exceeded 30%.
--- a/Configs/.local/lib/hyde/gpuinfo.sh
+++ b/Configs/.local/lib/hyde/gpuinfo.sh
@@ -194,11 +194,21 @@ general_query() {
     sensors_data=$(sensors 2>/dev/null)
     temperature=$(echo "$sensors_data" | $filter grep -m 1 -E "(edge|Package id.*|another keyword)" | awk -F ':' '{print int($2)}')
     fan_speed=$(echo "$sensors_data" | $filter grep -m 1 -E "fan[1-9]" | awk -F ':' '{print int($2)}')
-    for file in /sys/class/power_supply/BAT*/power_now; do
-        [[ -f $file ]] && power_discharge=$(awk '{print $1*10^-6 ""}' "$file") && break
+    local power_supply_dir="${GPUINFO_POWER_SUPPLY_DIR:-/sys/class/power_supply}"
+    local power_raw current_raw voltage_raw
+    # These attributes can exist and still fail to read — some laptops expose
+    # power_now but have their EC answer ENXIO for it — so read the value
+    # first and only use it if we actually got one. Letting awk open the file
+    # itself turns that into a fatal error on every poll.
+    for file in "$power_supply_dir"/BAT*/power_now; do
+        [[ -f $file ]] && power_raw=$(cat "$file" 2>/dev/null) && [[ -n $power_raw ]] &&
+            power_discharge=$(awk -v power="$power_raw" 'BEGIN {print power*10^-6 ""}') && break
     done
-    [[ -z $power_discharge ]] && for file in /sys/class/power_supply/BAT*/current_now; do
-        [[ -e $file ]] && power_discharge=$(awk -v current="$(cat "$file")" -v voltage="$(cat "${file/current_now/voltage_now}")" 'BEGIN {print (current * voltage) / 10^12 ""}') && break
+    [[ -z $power_discharge ]] && for file in "$power_supply_dir"/BAT*/current_now; do
+        [[ -e $file ]] && current_raw=$(cat "$file" 2>/dev/null) &&
+            voltage_raw=$(cat "${file/current_now/voltage_now}" 2>/dev/null) &&
+            [[ -n $current_raw && -n $voltage_raw ]] &&
+            power_discharge=$(awk -v current="$current_raw" -v voltage="$voltage_raw" 'BEGIN {print (current * voltage) / 10^12 ""}') && break
     done
     get_utilization() {
         statFile=$(head -1 /proc/stat)
--- a/tests/test_gpuinfo.sh
+++ b/tests/test_gpuinfo.sh
@@ -29,6 +29,7 @@ if [ -f "$state_file" ]; then
     cp -p "$state_file" "$state_backup"
 fi
 fake_bin=$(mktemp -d)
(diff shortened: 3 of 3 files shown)


cativo23:
While fixing the CI failure on this PR I found two more defects in the same function, and filed them separately as #2028 rather than growing this PR, since CONTRIBUTING.md asks to keep a pull request focused.

One of them is why the test assertion here is narrowed to `awk: fatal.*power_now` instead of matching any `awk: fatal`: `general_query()` also hands two cpufreq paths straight to awk, and those abort on a host with no cpufreq scaling driver, which is what the runner here has. That is the same defect class as this PR fixes, just a different call site. Matching broadly would have failed this case for a reason that has nothing to do with `power_now`.

This PR is unchanged in scope and still just fixes #2026.

kRHYME7:
merged. Please still observe for any regression/side effects.

Issue #1941, opened 2026-08-05
Issues on Hyprland v0.55+ Lua Migration: Gestures, Hyprlock, and HyDE Gallery Themes Conflict


nuzulachyar:
### Description

Hi team, I am using CachyOS with the latest Hyprland Lua migration (v0.55+). I encountered several broken features after transitioning, especially when applying themes from the HyDE Gallery:Touchpad Gestures: The 3-finger workspace swipe is broken/not working properly and fell back to 4-fingers constraint. Trying to override via userprefs.lua throws an unknown config key error for gestures.workspace_swipe.Hyprlock & Hypridle: Screen locking via hyprlock is broken because hl.dispatch.exec or hl.dsp.exec throws a nil value error in userprefs.lua. hypridle.conf also fails because it still uses the legacy .conf sourcing mechanism.HyDE Gallery Themes: Applying themes from the gallery causes a major breakdown, breaking Waybar stylesheets, fonts, and causing syntax loops because the theme files are still relying on the old .conf / Hyprlang structure.Environment:OS: CachyOSHyprland Version: v0.55+ (Lua config)

### Steps to Reproduce

1. Install CachyOS and update to the latest Hyprland version (v0.55+ with Lua configuration support).
2. Install HyDE Shell and attempt to download or apply a theme from the HyDE Gallery.
3. Observe that Waybar configurations, system fonts, and lock screen functions become broken due to old `.conf` files overriding the new `.lua` environment.
4. Try to configure touchpad gestures or map `hyprlock` manually inside `~/.config/hypr/userprefs.lua`.
5. Press `SUPER + L` or perform a 3-finger swipe on the touchpad.

### Expected Behavior

I expected the applied HyDE Gallery theme to adapt cleanly to the new Lua structure. I also expected that setting touchpad gestures (3-finger workspace swipe) and `hyprlock` inside `userprefs.lua` would function normally without throwing syntax or nil value errors, keeping the system stable.

### Actual Behavior

The system throws critical runtime errors. 
- Touchpad gestures fallback to a broken 4-finger constraint, and adding `gestures` to `userprefs.lua` throws an `unknown config key 'gestures.workspace_swipe'` error.
- Setting `hyprlock` with a lambda function throws an `attempt to call a nil value (field 'exec')` runtime error because the HyDE wrapper wrapper currently misses or misinterprets the dispatcher.
[...]

PR #2051, opened 2026-09-05
fix: swap swipe-gesture finger counts to 3=workspace, 4=focus

Delcado19:
## What

`dwindle`, `master`, `scrolling` and `scrolling-down` bound workspace-switch swipes to **4 fingers** and directional focus-switch to 3, while `monocle` used **3 fingers** for workspace-switch (with its own comment stating that intent: "Use 3-finger swipes to move between workspaces"). All five layouts came from the same Lua migration commit, so this was drift between copy-pasted files, not a deliberate per-layout difference.

This matches the near-universal desktop convention (GNOME/KDE etc. all use 3-finger swipe for workspace switching) and is exactly what's reported as broken in #1941 and one of the symptoms listed in #1945 ("Gestures are broken (e.g., 3-finger swipe not working, fallback to 4-finger or no response)").

**Note on scope:** this addresses only the gesture symptom from #1941/#1945 — both are multi-symptom reports and the rest of their listed issues are untouched by this PR.

## Fix

Swapped the finger counts in the four inconsistent layouts (workspace-switch → 3 fingers, directional focus → 4 fingers), matching `monocle.lua`.

## Test

Added `tests/test_layout_gestures.sh` (+ `tests/lua/layout_gestures_harness.lua`), which loads every shipped layout file against a stubbed Hyprland API and asserts every `action == "workspace"` gesture uses 3 fingers — so this can't silently drift apart again.

Also includes an unrelated small fix found while investigating: `tests/test_monitor_scale.sh`'s "unreadable scale" case wasn't hermetic — it queried the real running compositor when no explicit scale was given, so it passed or failed depending on whether the machine running the suite had a live Hyprland session (it failed here, since this machine's real scale is 1.5). The production fallback logic in `get_monitor_scale` was already correct; only the test needed a stubbed `hyprctl` to force the fallback path deterministically.

Verified live: deployed to a running HyDE session (Hyprland 0.56.2) and confirmed via `hyprctl reload` with no config errors. Full test suite passes (31/31).

Related: #1941, #1945 (partial)

## Summary by CodeRabbit

* **Bug Fixes**
  * Reassigned touchpad gestures across layouts so three-finger swipes switch workspaces, while four-finger swipes move focus.
  * Standardized horizontal and vertical gesture behavior across supported layouts.

* **Tests**
  * Added automated checks to detect incorrect or duplicate gesture bindings.
  * Improved monitor-scale testing by validating fallback behavior without requiring an active Hyprland session.

### DIFF
--- a/Configs/.local/share/hypr/lua/layouts/dwindle.lua
+++ b/Configs/.local/share/hypr/lua/layouts/dwindle.lua
@@ -29,9 +29,11 @@ hl.bind("ALT + U", hl.dsp.layout("movetoroot"), {description = "[Dwindle] move f
 hl.bind("ALT + Y", hl.dsp.layout("preselect l"), {description = "[Dwindle] preselect left/top split"})
 hl.bind("ALT + I", hl.dsp.layout("preselect r"), {description = "[Dwindle] preselect right/bottom split"})
 
+-- 3-finger swipes switch workspaces, matching the monocle layout and the
+-- near-universal desktop convention; 4-finger swipes move focus instead.
 hl.gesture(
     {
-        fingers = 4,
+        fingers = 3,
         direction = "horizontal",
         action = "workspace"
     }
@@ -39,7 +41,7 @@ hl.gesture(
 
 hl.gesture(
     {
-        fingers = 4,
+        fingers = 3,
         direction = "vertical",
         action = "workspace"
     }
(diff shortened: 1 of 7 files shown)


kRHYME7:
@Delcado19  Thanks! I don't review tests so please deal with it if anything borkes. Will merge now! 

I never thought of just using 3 for WS swipe and 4 for focus. Thanks! this is better and consistent!

PR #2056, opened 2026-09-06
fix: guard gpuinfo.sh's cpufreq reads and fix its cpuinfo_file typo (#2028)

### LINKED ISSUE #2028: [Bug]: two more spots in gpuinfo.sh where a missing path or an unset variable breaks general_query()
## Description

Two more defects in `general_query()` / `get_utilization()` in `Configs/.local/lib/hyde/gpuinfo.sh`, found while working on #2026. They're independent of each other and of #2026, and either can be fixed on its own.

**1. Two `awk` calls open sysfs paths directly and abort when they're missing** (lines 226-227 on `dev` @ `a1eb2c05`):

```sh
current_clock_speed=$(awk '{sum += $1; n++} END {if (n > 0) print sum / n / 1000 ""}' /sys/devices/system/cpu/cpufreq/policy*/scaling_cur_freq)
max_clock_speed=$(awk '{print $1/1000}' /sys/devices/system/cpu/cpu0/cpufreq/cpuinfo_max_freq)
```

This is the same shape as #2026: a path handed to `awk` with nothing checking it first. When there's no `cpufreq` in sysfs the glob doesn't match, bash passes the literal string through, and awk aborts:

```console
$ awk '{print}' /sys/devices/system/cpu/cpufreq/NOEXISTE*/scaling_cur_freq; echo "exit=$?"
awk: fatal: cannot open file `/sys/devices/system/cpu/cpufreq/NOEXISTE*/scaling_cur_freq' for reading: No such file or directory
exit=2
```

This project's own CI runner hits it. Both fatals showed up in the `Shell coverage` job on PR #2027 at commit `5477ddcf`:

```
awk: fatal: cannot open file `/sys/devices/system/cpu/cpufreq/policy*/scaling_cur_freq' for reading: No such file or directory
awk: fatal: cannot open file `/sys/devices/system/cpu/cpu0/cpufreq/cpuinfo_max_freq' for reading: No such file or directory
```

They no longer print on that PR's current green run, and it's worth being clear about why: they were never fixed, the test around them was narrowed. `tests/test_gpuinfo.sh` now matches `awk: fatal.*power_now` rather than any `awk: fatal`, precisely so an unrelated cpufreq failure wouldn't fail a case about `power_now`. The comment on that assertion says so, so the evidence is still in the tree even though the log has moved on.

What decides whether this bites is whether the host kernel has a cpufreq scaling driver loaded. Virtualized and cloud CPUs often don't (this project's runner among them), and neither do some ARM boards. A container inherits whatever its host has, so it isn't a separate case. Where it does bite, it's two fatal errors per poll at `"interval": 5`.

**2. `$cpuinfo_file` is never assigned** (lines 220-221), which makes a fallback write nowhere:

```sh
sed -i -e "..." "$gpuinfo_file" || {
    echo "GPUINFO_PREV_STAT=\"$currStat\"" >>"$cpuinfo_file"
    echo "GPUINFO_PREV_IDLE=\"$currIdle\"" >>"$cpuinfo_file"
}
```

`gpuinfo_file` is set at line 3; `cpuinfo_file` never is, anywhere in the script. It isn't invented, though: `Configs/.local/lib/hyde/cpuinfo.sh` sets `cpuinfo_file` at its own line 3 and carries the same block at lines 56-59, using it correctly. The `gpuinfo.sh` copy has the `sed -i` target renamed to `$gpuinfo_file` but the two fallback lines left pointing at the original name, so this looks like a copy of that pattern with the rename applied only partway. `hyde-shell` dispatches each script with `exec bash`, so nothing carries the variable across from a `cpuinfo` run either. The block only runs when the `sed -i` fails, so it's latent rather than constant, but when that path is taken the redirection expands to an empty filename:

```console
$ unset cpuinfo_file; echo test >>"$cpuinfo_file"
bash: : No such file or directory
```

The utilization state then isn't persisted, which is exactly what the fallback exists to prevent. `shellcheck` already flags it as SC2154.

## Steps to Reproduce

For (1), on any machine or container without `/sys/devices/system/cpu/cpufreq`:

```console
$ hyde-shell gpuinfo 2>&1 >/dev/null
awk: fatal: cannot open file `/sys/devices/system/cpu/cpufreq/policy*/scaling_cur_freq' for reading: No such file or directory
awk: fatal: cannot open file `/sys/devices/system/cpu/cpu0/cpufreq/cpuinfo_max_freq' for reading: No such file or directory
```

For (2):

```console
$ grep -n 'cpuinfo_file' Configs/.local/lib/hyde/gpuinfo.sh
220:            echo "GPUINFO_PREV_STAT=\"$currStat\"" >>"$cpuinfo_file"
221:            echo "GPUINFO_PREV_IDLE=\"$currIdle\"" >>"$cpuinfo_file"
$ shellcheck Configs/.local/lib/hyde/gpuinfo.sh
... SC2154 (warning): cpuinfo_file is referenced but not assigned (did you mean 'gpuinfo_file'?)
```

## Expected Behavior

(1) A sysfs path that isn't there gets skipped, and the tooltip omits clock speed, the same way it already behaves when a value is genuinely unavailable.

(2) The fallback appends to the same state file the `sed -i` was targeting.

## Actual Behavior

(1) Two `awk` fatal errors on stderr per poll, forever, on any machine without cpufreq.

(2) The fallback writes to an empty filename, bash reports `: No such file or directory`, and the utilization state is lost for that cycle.

## Environment

Both are visible by reading the source on `dev` @ `a1eb2c05`, and (1) reproduces in this project's own CI container, so neither is specific to my hardware. For completeness, the machine I found them on:

- Graphics Driver: intel (TigerLake-LP GT2, i915)
- Display Server: Wayland
- HyDE: present on both `master` (`7c7b832d`) and `dev` (`a1eb2c05`)

## Additional Information

I'm keeping these out of #2027 deliberately, since CONTRIBUTING.md asks to keep a pull request focused, and #2027 is scoped to `power_now`. I'll open a PR for either or both once #2027 lands, unless you'd rather handle them differently.

Delcado19:
Fixes #2028.

## What was wrong

1. `general_query()` let `awk` open `/sys/devices/system/cpu/cpufreq/policy*/scaling_cur_freq` and `.../cpu0/cpufreq/cpuinfo_max_freq` directly. On any host with no cpufreq scaling driver loaded (virtualized/cloud CPUs, some ARM boards, this project's own CI runner), the glob doesn't match, bash passes the literal pattern through, and awk fatals -- two stderr errors on every single poll, forever. Same shape as the `power_now` fix in #2027.

2. `get_utilization()`'s `sed -i` failure fallback wrote to `$cpuinfo_file`, a variable `gpuinfo.sh` never assigns (a copy-paste leftover from `cpuinfo.sh`, which sets it and uses it correctly). The fallback silently wrote to an empty filename instead of persisting utilization state.

## Fix

- Read each cpufreq file only if it exists, and let `awk` do the averaging over the ones that do -- same pattern already used for `power_now`/`current_now` just above it in the file.
- Fixed the fallback to use `$gpuinfo_file`.

## A regression caught along the way

My first pass summed the readings with bash's own `$((...))` arithmetic. That's fragile against exactly the kind of out-of-spec content this fix is meant to be defensive about: anything that isn't a plain integer (e.g. a stray decimal point) makes bash throw its own `arithmetic syntax error` and print a diagnostic to stderr on every poll -- the same class of noise this issue is about removing, just from a different builtin. Switched the summation to `awk`, which coerces malformed content to `0` instead, matching how the rest of the function already treats unreadable/missing sysfs values.

## Tests

- Missing cpufreq tree entirely (the actual bug repro: no fatal, valid JSON).
- Well-formed multi-policy readings average to the correct value (not just "doesn't crash").
- Malformed (non-integer) content degrades gracefully instead of leaking an arithmetic error -- pinned to `LC_ALL=C.utf8` so the assertion doesn't depend on the host's locale (bash localizes that message; this dev machine has it in German).
- Static guard that the utilization fallback no longer references `$cpuinfo_file`.

All 32 existing test cases still pass.

## Summary by CodeRabbit

- **Bug Fixes**
  - GPU information collection no longer produces fatal errors on systems without CPU frequency data.
  - CPU clock-speed reporting safely handles missing, incomplete, or malformed frequency readings.
  - Utilization fallback handling records data correctly and avoids invalid output paths.
  - GPU information continues to emit valid JSON when hardware metrics are unavailable or contain unexpected values.
  - Available CPU frequency data is averaged and reported accurately when multiple readings are present.
  - CPU temperature detection now supports additional AMD sensor labels, including Tctl and Tdie.

### DIFF
--- a/Configs/.local/lib/hyde/gpuinfo.sh
+++ b/Configs/.local/lib/hyde/gpuinfo.sh
@@ -238,15 +238,18 @@ general_query() {
     # /sys/devices/system/cpu/cpufreq tree at all, so the glob doesn't match and
     # bash passes the literal pattern through -- letting awk open it directly
     # turns that into a fatal error on every poll (#2028). Read each file only
-    # if it exists, and feed the values to awk instead of the path.
+    # if it exists, and let awk do the summing: bash's own $((...)) throws a
+    # syntax error (killing the whole script, not just this calculation) on
+    # anything that isn't a plain integer, whereas awk coerces malformed sysfs
+    # content to 0 instead of dying on it.
     local cpu_sysfs_dir="${GPUINFO_CPU_SYSFS_DIR:-/sys/devices/system/cpu}"
-    local clock_sum=0 clock_n=0 freq
+    local freq clock_freqs=()
     for file in "$cpu_sysfs_dir"/cpufreq/policy*/scaling_cur_freq; do
-        [[ -f $file ]] && freq=$(cat "$file" 2>/dev/null) && [[ -n $freq ]] &&
-            clock_sum=$((clock_sum + freq)) && clock_n=$((clock_n + 1))
+        [[ -f $file ]] && freq=$(cat "$file" 2>/dev/null) && [[ -n $freq ]] && clock_freqs+=("$freq")
     done
     current_clock_speed=""
-    [[ $clock_n -gt 0 ]] && current_clock_speed=$(awk -v sum="$clock_sum" -v n="$clock_n" 'BEGIN {print sum / n / 1000 ""}')
+    ((${#clock_freqs[@]} > 0)) &&
+        current_clock_speed=$(printf '%s\n' "${clock_freqs[@]}" | awk '{sum += $1; n++} END {if (n > 0) print sum / n / 1000 ""}')
     max_clock_speed=""
     max_freq_file="$cpu_sysfs_dir/cpu0/cpufreq/cpuinfo_max_freq"
     if [[ -f $max_freq_file ]]; then
--- a/Configs/.local/lib/hyde/gpuinfo.sh
+++ b/Configs/.local/lib/hyde/gpuinfo.sh
@@ -227,14 +227,32 @@ general_query() {
         GPUINFO_PREV_STAT=$currStat
         GPUINFO_PREV_IDLE=$currIdle
         sed -i -e "/^GPUINFO_PREV_STAT=/c\GPUINFO_PREV_STAT=\"$currStat\"" -e "/^GPUINFO_PREV_IDLE=/c\GPUINFO_PREV_IDLE=\"$currIdle\"" "$gpuinfo_file" || {
-            echo "GPUINFO_PREV_STAT=\"$currStat\"" >>"$cpuinfo_file"
-            echo "GPUINFO_PREV_IDLE=\"$currIdle\"" >>"$cpuinfo_file"
+            echo "GPUINFO_PREV_STAT=\"$currStat\"" >>"$gpuinfo_file"
+            echo "GPUINFO_PREV_IDLE=\"$currIdle\"" >>"$gpuinfo_file"
         }
         awk -v stat="$diffStat" -v idle="$diffIdle" 'BEGIN {total=stat+idle; printf "%.1f", (total > 0 ? (stat/total)*100 : 0)}'
     }
     utilization=$(get_utilization)
-    current_clock_speed=$(awk '{sum += $1; n++} END {if (n > 0) print sum / n / 1000 ""}' /sys/devices/system/cpu/cpufreq/policy*/scaling_cur_freq)
-    max_clock_speed=$(awk '{print $1/1000}' /sys/devices/system/cpu/cpu0/cpufreq/cpuinfo_max_freq)
+    # As with power_now above: a host with no cpufreq scaling driver (common on
+    # virtualized/cloud CPUs and this project's own CI runner) has no
+    # /sys/devices/system/cpu/cpufreq tree at all, so the glob doesn't match and
+    # bash passes the literal pattern through -- letting awk open it directly
+    # turns that into a fatal error on every poll (#2028). Read each file only
+    # if it exists, and feed the values to awk instead of the path.
+    local cpu_sysfs_dir="${GPUINFO_CPU_SYSFS_DIR:-/sys/devices/system/cpu}"
+    local clock_sum=0 clock_n=0 freq
+    for file in "$cpu_sysfs_dir"/cpufreq/policy*/scaling_cur_freq; do
+        [[ -f $file ]] && freq=$(cat "$file" 2>/dev/null) && [[ -n $freq ]] &&
+            clock_sum=$((clock_sum + freq)) && clock_n=$((clock_n + 1))
+    done
+    current_clock_speed=""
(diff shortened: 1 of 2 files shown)


Delcado19:
Re: the "outside diff" CodeRabbit findings on this PR:

**"Remove the duplicate script invocations"** — checked empirically rather than assuming: there's exactly one `bash "$script"` call in each of those fixtures (lines 250/273/300/339 all assign to a different single variable each). The two JSON lines in the captured output come from `gpuinfo.sh` itself, not from a second invocation: on a cold start (state file absent), the script's own startup-detection logic re-invokes `general_query()` a second time within the *same* process, which is why `tests/test_gpuinfo.sh`'s very first comment already documents this ("the script legitimately emits more than one object per cold-start run, because the startup detection re-invokes it"). Confirmed again just now: a single manual `bash gpuinfo.sh` run with the same stubbed `sensors` reliably prints 2 JSON lines. Nothing to remove here.

**"Assert that absent cpufreq data omits clock-speed output"** — fair point, added (checks the tooltip has no "Clock Speed" segment when `GPUINFO_CPU_SYSFS_DIR` has no cpufreq tree).

**"Make the utilization fallback assertion behavioral"** — agreed this would be stronger, but `gpuinfo_file` is hardcoded to `/tmp/hyde-$UID-gpuinfo` with no override, unlike `GPUINFO_POWER_SUPPLY_DIR`/`GPUINFO_CPU_SYSFS_DIR` elsewhere in this same script. Reliably forcing the `sed -i` in `get_utilization()` to fail (while leaving the fallback's own `>>` append still writable, to actually observe the fix) needs a read-only *directory* with a writable file inside it (verified: `sed -i`'s rename-based in-place edit fails on a read-only dir even when the file itself is writable, while a plain `>>` append to that same file still succeeds) -- and I don't have a way to point `gpuinfo_file` at a test-controlled directory without changing the production script just to make this easier to test, which felt like a step too far for this PR. Kept the static guard (grep for the old wrong variable name) as-is.

kRHYME7:
cc @rubiin

Delcado19:
Superseded by #2062 — `gpuinfo.sh` no longer exists after that PR (rewritten in Lua per kRHYME7's request on #2058), so this cpufreq guard has no file left to apply to. The Lua clock-speed reader in #2062 was written fresh with the same guarded-existence-check approach from the start, so the fix isn't lost, just carried over in the new file.

Suggest closing this one once #2062 lands.

rubiin (approved):

PR #2059, opened 2026-09-06
fix: source globalcontrol.sh in hyde-shell's app dispatch (#1951)

### LINKED ISSUE #1951: [Bug]: battery.notify config not propagated to batterynotify.lua after Lua migration
### Description

After updating HyDE to the latest version, the battery notification daemon appears to no longer receive the settings from `~/.config/hyde/config.toml`.

I have the following configuration:

```toml
[battery.notify]
dock = true
```

This worked before the recent update and prevented charger plug-in/plug-out and battery-full notifications while using the laptop docked.

After updating, I now receive frequent notifications such as:

* `Charger Plug Out`
* `Battery Full`
* charger state change notifications

even though `dock = true` is still present under `[battery.notify]`.

The generated HyDE environment file correctly contains:

```bash
export BATTERY_NOTIFY_DOCK=true
```

in:

```text
~/.local/state/hyde/config
```

However, the running `batterynotify.lua` process does not have `BATTERY_NOTIFY_DOCK` in its environment.

### Steps to Reproduce

1. Set:

```toml
[battery.notify]
dock = true
```

in:

```text
~/.config/hyde/config.toml
```

2. Confirm HyDE correctly generates the variable:

```bash
grep '^export BATTERY_NOTIFY' ~/.local/state/hyde/config
```

Output includes:

```text
export BATTERY_NOTIFY_DOCK=true
```

3. Check the environment of the running battery notifier:

```bash
pid=$(pgrep -n -f batterynotify.lua)
tr '\0' '\n' < /proc/$pid/environ | grep '^BATTERY_NOTIFY'
```

No `BATTERY_NOTIFY_*` variables are present.

4. Plug/unplug the charger or leave the system docked near/full battery level.

Battery state notifications are shown despite `dock = true`.

### Expected Behavior

The `batterynotify.lua` service should inherit the configuration generated from `~/.config/hyde/config.toml`.

With:

```toml
[battery.notify]
dock = true
```

the daemon should receive:

```text
BATTERY_NOTIFY_DOCK=true
```

and suppress charger status-change / battery-full notifications as intended.

### Actual Behavior

The generated `~/.local/state/hyde/config` contains the correct variable, but the running `batterynotify.lua` service does not receive it.

As a result, `batterynotify.lua` falls back to its default:

```lua
dock = getenv_bool('BATTERY_NOTIFY_DOCK', false)
```

and behaves as if dock mode were disabled.

### Screenshots

_No response_

### Environment

```text
Date: Wed Aug 5 14:13:21 2026
Tag: v0.56.2, commits: 7661

Libraries:
Hyprutils: built against 0.14.0, system has 0.14.0

no flags were set


System Information:
System name: Linux
Node name: ironasus
Release: 7.1.6-arch1-1
Version: #1 SMP PREEMPT_DYNAMIC Tue, 04 Aug 2026 11:19:27 +0000

Libraries:
Hyprutils: built against 0.14.0, system has 0.14.0



GPU information: 
00:02.0 VGA compatible controller [0300]: Intel Corporation Kaby Lake-R GT2 [UHD Graphics 620] [8086:5917] (rev 07) (prog-if 00 [VGA controller])
01:00.0 3D controller [0302]: NVIDIA Corporation GM108M [GeForce MX130] [10de:174d] (rev a2)
NVRM version: NVIDIA UNIX x86_64 Kernel Module  580.173.02  Tue Jun 23 08:38:17 UTC 2026


os-release: NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=arch
BUILD_ID=rolling
ANSI_COLOR="38;2;23;147;209"
LOGO=archlinux-logo


plugins:

Explicit sync: supported
GL ver: 3.2
Backend: drm

Monitor info:
	Panel eDP-1: 1920x1080, eDP-1 LG Display 0x0573  -> backend drm
		explicit ✔️
		edid:
			hdr ❌
			chroma ✔️
			bt2020 ❌
		vrr capable ❌
		non-desktop ❌
		

State:

configProvider: lua
backend: drm




======Config-Start======
Not supported under lua
======Config-End========
```

## Suspected problem

This appears to be related to the new Lua/service startup path.

The battery notifier is started from `variables.lua` through:

```text
hyde-shell app ... -- batterynotify.lua
```

In `hyde-shell`, the `app)` case executes `app.sh` before `globalcontrol.sh` is sourced.

`globalcontrol.sh` is where `~/.local/state/hyde/config` is normally sourced.

So services launched through the early `hyde-shell app` path may not receive config variables generated from `config.toml`.

This may affect other services/configuration values launched through the same path as well, not only `BATTERY_NOTIFY_DOCK`.

The issue appeared after updating across the recent shell → Lua migration; the previous battery notifier behavior worked correctly.

Delcado19:
Fixes #1951.

## What was wrong

Every `hc.start.*` entry in `variables.lua` launches its service through `hyde-shell app -u ... -t service|scope -- <cmd>`. That subcommand is handled by its own early `case "$1" in app) ... exec app.sh "$@" ;; ... esac` branch, which `exec`s straight into `app.sh` (and from there into `app2unit`, or a plain `exec` when there's no systemd user session) -- it never reaches `source "${LIB_DIR}/hyde/globalcontrol.sh"` further down the script, which every *other* hyde-shell subcommand goes through.

`globalcontrol.sh` calls `export_hyde_config()`, which sources `~/.local/state/hyde/config` -- the env vars HyDE generates from `~/.config/hyde/config.toml` (`BATTERY_NOTIFY_DOCK`, in the report). So a service launched via `app` never saw those variables: the generated file had the right value, but the running process's environment didn't, and it silently fell back to its own default.

This affects any `hc.start.*` service, not just `batterynotify.lua` -- `dunst`, the applets, `hyprsunset`, etc. all launch through the same `app` path.

## Fix

Source `globalcontrol.sh` inside the `app)` case too, before the `exec`, matching the error handling already used on the other path. `export`ed variables set before an `exec` are inherited by the process that replaces the shell, so this reaches the actual launched service.

## Tests

New `tests/test_hyde_shell_app_env.sh`:
- a sandboxed `HOME` with a generated `~/.local/state/hyde/config` setting `BATTERY_NOTIFY_DOCK=true` -- confirms `hyde-shell app -- <cmd>` now forwards it to the launched process (fails against the pre-fix script, verified locally).
- no generated config file at all (a fresh install) -- confirms this doesn't turn into a hard failure, since `export_hyde_config()` already treats a missing file as success and that has to keep holding on this newly added call site too.

Both a real systemd user session and its absence hit the same code path in this test (a stubbed `app2unit` stands in for either), so it's not dependent on the runner's systemd state.

All existing test cases still pass (33/33).

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed `hyde-shell app` so generated configuration environment variables are available to app-managed services.
  - Added clear error handling when configuration loading fails.
  - Commands continue to work when no generated configuration file is present.

- **Tests**
  - Added coverage verifying environment variable forwarding and behavior with missing configuration.

### DIFF
--- a/Configs/.local/bin/hyde-shell
+++ b/Configs/.local/bin/hyde-shell
@@ -516,6 +516,16 @@ export BIN_DIR LIB_DIR SHARE_DIR PATH HYDE_SCRIPTS_PATH
 
 case "$1" in
 app)
+    # app launches a service/scope that outlives this shell via exec, so
+    # anything it needs from HyDE's generated config (e.g. config.toml ->
+    # env vars like BATTERY_NOTIFY_DOCK) has to be sourced and exported
+    # before the exec below -- this path used to skip globalcontrol.sh
+    # entirely and reach it only on the fallback branch further down,
+    # silently launching every app-managed service without it (#1951).
+    if ! source "${LIB_DIR}/hyde/globalcontrol.sh"; then
+        echo "Error: Could not load HyDE, broken installation?"
+        exit 1
+    fi
     export PATH="${HYDE_SCRIPTS_PATH}:${PATH}"
     [[ "${HYDEPY}" -eq 1 ]] && python_activate
     shift
--- a/tests/test_hyde_shell_app_env.sh
+++ b/tests/test_hyde_shell_app_env.sh
@@ -1,0 +1,84 @@
+#!/usr/bin/env bash
+# `hyde-shell app` launches a service/scope via `exec`, so a process started
+# through it (every hc.start.* entry in variables.lua uses this path) has to
+# inherit whatever HyDE's generated config exports. This case used to skip
+# globalcontrol.sh entirely -- and with it export_hyde_config(), which
+# sources ~/.local/state/hyde/config (config.toml's generated env vars, e.g.
+# BATTERY_NOTIFY_DOCK) -- and only reached it on hyde-shell's other,
+# non-"app" code path (#1951).
+
+. "$(dirname -- "$0")/lib/common.sh"
+
+hyde_shell="$REPO_ROOT/Configs/.local/bin/hyde-shell"
+[ -f "$hyde_shell" ] || {
+    fail "hyde-shell not found at $hyde_shell"
+    finish
+}
+
+work_dir=$(mktemp -d)
+trap 'rm -rf "$work_dir"' EXIT
+
+home_dir="$work_dir/home"
+mkdir -p "$home_dir/.local/state/hyde" "$home_dir/.config"
+printf 'export BATTERY_NOTIFY_DOCK=true\n' >"$home_dir/.local/state/hyde/config"
+
+# app.sh hands off to app2unit when a systemd user session exists, or to a
+# plain `exec` of the trailing command otherwise -- this fake stands in for
+# both identically, so the test doesn't depend on whether this machine
+# happens to have one running.
+fake_bin="$work_dir/fake_bin"
+mkdir -p "$fake_bin"
+cat >"$fake_bin/app2unit" <<'EOF'
+#!/bin/sh
+while [ "$#" -gt 0 ] && [ "$1" != "--" ]; do shift; done
+[ "$#" -gt 0 ] && shift
+exec "$@"
+EOF
+chmod +x "$fake_bin/app2unit"
+
+bin_dir=$(dirname "$hyde_shell")
+
+##
+# Runs `hyde-shell app -- "$@"` against the sandboxed HOME, with the fake
+# app2unit and the real hyde-shell/its sibling scripts on PATH.
+##
+run_app() {
+    env -i \
+        HOME="$home_dir" \
+        XDG_CONFIG_HOME="$home_dir/.config" \
+        XDG_DATA_HOME="$home_dir/.local/share" \
+        XDG_CACHE_HOME="$home_dir/.cache" \
+        XDG_STATE_HOME="$home_dir/.local/state" \
+        XDG_RUNTIME_DIR="$home_dir/run" \
+        PATH="$fake_bin:$bin_dir:/usr/bin:/bin" \
+        bash "$hyde_shell" app -- "$@"
+}
+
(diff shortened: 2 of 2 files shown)


Configs/.local/bin/hyde-shell
kRHYME7, line 532: Is it possible to source it inside the app.sh???

kRHYME7 (changes requested):

Delcado19:
Checked this rather than just guessing: `app.sh` is called directly (not just through `hyde-shell app`) from `lockscreen.sh`, `resetxdgportal.sh`, `rofilaunch.sh`, `wallpaper.hyprpaper.sh`, and `hyprlock.sh`. Every one of those already does `eval "$(hyde-shell init)")` (or, for `hyprlock.sh`, sources `hyde-shell` directly) as its first executable line -- which itself runs `globalcontrol.sh` (`hyde-shell init`'s `initialized()` literally `cat`s the whole file for the caller to `eval`). So by the time any of them reaches its own `app.sh ...` call, `export_hyde_config()` has already run in that process.

Sourcing it again inside `app.sh` would re-run that for every one of those callers on every invocation -- and `rofilaunch.sh`'s `-run-command "app.sh --fuzzel-compat -- {cmd}"` means that's *every app launch from the menu*, not an occasional path. `globalcontrol.sh` does real work (theme detection, wallbash dir setup, etc.), so that's repeated overhead on an already-hot path for callers that don't need it.

The actual gap this PR fixes is narrower: `hc.start.*` in `variables.lua` calls `hyde-shell app -u ... -- <cmd>` directly from Hyprland's own `hl.exec_cmd(...)`, with no prior `eval "$(hyde-shell init)")` step -- that's the one path that was actually missing it, and it's specific to `hyde-shell`'s own `app)` dispatch, not `app.sh` itself.

(`app.sh`'s `#!/usr/bin/env sh` shebang turned out not to be a blocker either way -- on Arch, `/bin/sh` is bash, and bash-as-`sh` still accepts `[[ ]]`/`declare -A` in my testing -- but the redundant re-sourcing on every app launch seemed like the stronger reason to keep this in `hyde-shell` rather than move it.)

PR #2062, opened 2026-09-06
feat(gpuinfo): rewrite gpuinfo.sh in Lua

Delcado19:
## Why

kRHYME7 asked for this directly on #2058:

> Can you propose a lua rewrite on this script? Too much sys calls by calling non bash native executables. This bash script was my attempt to make a Cross-gpu script and this is my best effort to make it efficient. And I think I already hit the limit for bash. (still being maintainable) I think using lua for calculations and text parsing will be more efficient and maintainable for future references.

`gpuinfo.sh` had grown into `lspci | grep` scans, `awk` field math, `sed -i` in-place state edits, and regex-based `sensors` text parsing, run on every waybar poll. This replaces it with `Configs/.local/lib/hyde/gpuinfo.lua`, one file in the same style as `altab.lua`/`batterynotify.lua`/`dconf.lua`: a requireable module of small functions plus a guarded CLI entry point, so every function is independently unit-testable without invoking the CLI.

## Why this is not a "zero subprocess calls" rewrite

Checked prior art before designing this (AwesomeWM's `lain`/`vicious` widgets — mature, years-old Lua libraries for exactly this problem): both still shell out (`find`, `sensors`) rather than reimplement lm-sensors' chip/label database. NVIDIA temperature/utilization/clock data is only available via `nvidia-smi` regardless of language — there is no sysfs/hwmon path for the proprietary driver. Reimplementing `lspci`'s PCI-ID name database in Lua would mean vendoring `pci.ids` and its parsing rules for a problem `lspci` already solves correctly.

So the goal was **fewer, more targeted subprocess calls**, with Lua replacing every shell pipeline that doesn't strictly need an external tool — not eliminating `sensors`/`nvidia-smi`/`lspci` outright.

## What actually changed

| Concern | Bash (`gpuinfo.sh`) | Lua (`gpuinfo.lua`) |
|---|---|---|
| GPU vendor detection | 3 separate full-bus `lspci -nn \| grep` scans, every poll | one pure-Lua `/sys/bus/pci/devices/*/class`+`vendor` scan; **one** targeted `lspci -nn -s <addr>` call, only for the human-readable name, only on first detection (cached in state afterward) |
| nouveau / `nvidia-smi` presence | `lsmod \| grep nouveau`, `command -v nvidia-smi` | `/proc/modules` read + `PATH` search, both pure Lua |
| Temperature + fan speed | `sensors` plaintext piped through `grep -m1 -E "..." \| awk` | one `sensors -j` call, parsed with the existing `luautils.json` decoder (same pattern `cpuinfo.sh` already uses, Lua instead of Perl) |
| Battery discharge | `awk` arithmetic over `cat`'d sysfs values | pure Lua (`tonumber`, no subprocess) |
| CPU utilization | `awk` field sums + `sed -i` in-place state file edit | pure Lua `/proc/stat` parsing + JSON state rewrite |
| CPU clock speed | `awk` averaging over a cpufreq glob | pure Lua (`lfs.dir` + `io.open`, guarded existence checks) |
| State persistence | hand-rolled `KEY="value"` shell file, incremental `sed -i` edits | one JSON object via `luautils.json`, read-mutate-write |
| CLI parsing | hand-rolled `case "$1" in` | `luautils.argparse` (already used by `altab.lua`/`config.lua`/`shaders.lua`) |
| AMD / NVIDIA branches | `amdgpu.py` (kept) → `jq`+`sed`; `nvidia-smi` → `IFS=',' read -ra` | same external calls (unavoidable, see above), parsed with `luautils.json.decode` / Lua `gmatch` |

Net effect on this Intel machine specifically (verified live, see Testing): 2 subprocess calls per poll (`sensors -j`, and `lspci` only once at first detection) instead of the bash version's `lspci` scans + `awk`/`sed` chain every single poll.

## Documented behavior change

Sensor-match precedence: the old `grep -m1 -E "(edge|Package id.*|Tctl|Tdie)"` picked whichever label came first in `sensors`' own arbitrary plaintext order. `sensors -j` decoded in Lua has no defined key order, so "first in the dump" can't be preserved without extra bookkeeping for an already-accidental property. Replaced with an explicit priority: `edge`/`junction` (GPU) → `Tctl`/`Tdie` (AMD CPU) → `Package id N` (Intel CPU) — GPU temperature always wins over a CPU proxy reading when both exist, which is what a module named "gpuinfo" should prefer anyway.

## Fixed along the way

A consolidated review pass caught and fixed several correctness bugs before this was considered done (see individual commits): toggle availability was reading the wrong state key (could select a vendor that isn't actually present), `nvidia-smi`'s literal `"[N/A]"` could crash JSON assembly, `--tired` never actually detected suspend due to a duplicated PCI domain prefix, and a couple of others.

Also fixed after review of a live run on real hardware: `lookup_pci_name` originally either stripped every `[...]` bracket (matching old bash behavior, but losing marketing names like `[HD Graphics 620]`) or too little (leaking AMD's own `[AMD/ATI]` vendor-alias bracket from `pci.ids`, which would show as `AMD [AMD/ATI] Navi 23 [...]`). Settled on: strip only the trailing `[vendor:device]` id bracket and AMD's specific alias bracket, keep everything else — so the displayed name is `Intel Kaby Lake-U GT2 [HD Graphics 620]`, `NVIDIA GA106 [GeForce RTX 3060 ...]`, `AMD Navi 23 [Radeon RX 6600 ...]`.

## Testing

- Full new unit test suite (`tests/lua/gpuinfo_*_spec.lua` + `tests/test_gpuinfo_lua_*.sh`), one spec per concern (state, vendor detection, toggle, battery, utilization, clock speed, sensors, NVIDIA, AMD, CLI, output assembly), plus an end-to-end regression port of every case the old `tests/test_gpuinfo.sh` covered.
- `bash tests/run.sh`: 44/44 passing.
- Live-verified against this machine's real hardware (Intel HD 620 / Kaby Lake-U) — cold + warm poll, `--stat`, `--toggle`, all exit 0, output cross-checked against the still-installed bash version for consistency.
- No NVIDIA/AMD hardware available to test against live; those branches and the name-parsing edge cases are covered by fixtures built from real `sensors -j` output shapes and real `pci.ids` vendor/device entries, not guessed strings.

## Also in this PR

- Waybar wiring updated so `custom-gpuinfo#nvidia.jsonc` picks up the `.lua` file the same way the other 3 gpuinfo modules already do (`hyde-shell`'s resolver prefers `.lua` over `.sh` automatically for those).
- `Configs/.local/lib/hyde/gpuinfo.sh` and `tests/test_gpuinfo.sh` removed.

## Supersedes #2056

#2056 (guard `gpuinfo.sh`'s cpufreq reads) has no file left to apply to once this merges — the Lua clock-speed reader was written fresh with guarded existence checks from the start. Recommend closing it once this lands; happy to leave a note there myself once this is confirmed.

## Summary by CodeRabbit

* **New Features**
  * Waybar GPU information now supports NVIDIA, AMD, Intel, and fallback readings with vendor selection, persistent state, temperature, utilization, clock speed, power, fan, tooltip, and icon reporting.
  * GPU temperature detection prioritizes GPU readings over CPU-proxy sensors.

* **Bug Fixes**
  * Missing or malformed hardware data now falls back safely while maintaining valid output.
  * Improved window border colors for windows that cannot join groups.
  * Dropdown terminal animation settings now persist across configuration reloads.
  * Idle screen power-off and resume behavior is more reliable.

* **Tests**
  * Expanded coverage for GPU vendors, sensors, power, state, CLI behavior, output, and fallback scenarios.


kRHYME7:
this MR should be Ack by other users with different hardware.

kRHYME7:
BTW is it possible for the test to be on a different repo? or treat it as sub module? This way you can keep test balloon but users who just use HyDE can download a leaner version of HyDE.Devs can just pull the submodule.

But id Don't know if github actions can handle it too.

Delcado19:
What's the main driver here for you -- download/clone size, keeping the main repo's contribution surface simpler for non-dev users, something about how the repo presents on GitHub, or something else (e.g. packaging tests separately for downstream distros)? Trying to weigh this against the actual cost before committing to it.

From a technical standpoint: install.sh never reads or deploys tests/ to the end user either way, so nothing changes for what actually lands on someone's machine, submodule or not. The real cost would be on the contributor side -- right now a fix and its regression test land in one PR (the pattern this whole repo's history uses), and a submodule would split that across two repos and two PRs for every change touching both. Also unresolved: whether GitHub Actions handles a submodule cleanly here, which you flagged yourself as uncertain.

Happy to help sort out the CI side if you want to pursue it, once we're clear on what the main goal is.

kRHYME7:
Never tried it by I thought we can put test as submodule so changes on test is handled on that separate repo. Same goes to the hyde site. 

This way normal users can  just pull the dotfiles and we can harden the test or adding more guardrails without ballooning the main repo. 

Another selfish reason is I can't invest more time reviewing on a thousand line test when the change is small.

I really think this is possible and CI supports it. If I have some time I'll try to dig if CI supports submodules.
Note HyDE will hold the config for CI, I think we can make it pull or use  a submodule. These are my speculations only.

Delcado19:
I ran the current PR branch against a real NVIDIA GPU and the live NVIDIA path works end to end.

Validation completed successfully:

- `gpuinfo.lua --use nvidia --startup` emitted valid Waybar JSON and populated the NVIDIA tooltip
- `gpuinfo.lua --stat nvidia` reported `nvidia_enable: true`
- automatic invocation selected the NVIDIA path and returned valid output
- `bash tests/test_gpuinfo_lua_nvidia.sh`
- `bash tests/test_gpuinfo_lua_vendor.sh`
- `bash tests/test_gpuinfo_lua_output.sh`
- `bash tests/test_gpuinfo_lua_e2e.sh`

The live values matched the direct NVIDIA query during the test run.

Delcado19:
Closing this PR for a substantial rework. We did not account for several hardware and driver differences. In particular, parts of the Lua implementation were copied 1:1 from gpuinfo.sh even though some of those paths can return incorrect or unrelated values, including fallback data. As a result, the current implementation does not meet the requirement to report only real values from the selected GPU. We will rework the vendor-specific data sources, remove invalid fallbacks, and add coverage for unsupported and invalid measurements before opening an updated PR.

PR #2071, opened 2026-09-08
fix: give swaync's notification window a transparent GTK background

### LINKED ISSUE #1611: [Bug]: Black background on right side when swaync 
  notification appears
# [Bug]: Black background appears on right side of screen when swaync notification pops up

## Description

When a notification fires, a black rectangle appears on the right side of the screen covering the full height of the monitor. This is caused by two issues in HyDE's default configuration:

1. **`Configs/.config/hypr/windowrules.conf`** applies `blur true` + `ignore_alpha 0` to `swaync-notification-window`. Because swaync's layer-shell surface spans the **full monitor height** (due to `fit-to-screen: true`), Hyprland renders blur across the entire 400×(monitor height) transparent surface. `ignore_alpha 0` only excludes pixels with *exactly* 0 alpha, so any anti-aliasing or partial transparency still gets blurred into a dark artifact.

2. **`Configs/.local/share/hyde/wallbash/theme/swaync.dcol`** (which generates `theme.css`) is missing `notificationwindow, blankwindow { background: transparent; }`. Without this rule, the GTK window node for swaync keeps its default dark background, making the entire layer surface appear black.

## Steps to Reproduce

1. Fresh HyDE install with default `windowrules.conf` and swaync config
2. Trigger any notification (e.g. `notify-send "Test" "Hello"`)
3. Observe a black rectangle on the right side of the screen for the duration of the notification

Easier to reproduce on a multi-monitor setup where one monitor is portrait (tall), because the layer surface is even taller.

## Expected Behavior

Only the notification popup itself is visible; the rest of the screen is unaffected.

## Actual Behavior

A black/dark rectangle (400px wide, full monitor height) appears on the right side of the screen whenever a notification is active.

### Issue 1 — Missing GTK transparency rule in swaync CSS template

`swaync.dcol` generates `theme.css` but does not include:

```css
notificationwindow, blankwindow {
    background: transparent;
    background-color: transparent;
}
```

This is the correct GTK CSS node name for the swaync notification window (confirmed in swaync's own default CSS at `/etc/xdg/swaync/style.css`). Without it, GTK renders a dark background across the full layer-shell surface.

### Issue 2 — `ignore_alpha 0` too low for blur on swaync-notification-window

In `windowrules.conf`:
```
layerrule = blur true,match:namespace swaync-notification-window
layerrule = ignore_alpha 0,match:namespace swaync-notification-window
```

`ignore_alpha 0` means only fully transparent pixels (alpha = 0.0 exactly) are excluded from blur. Any pixel with even slight alpha (from GTK rendering/anti-aliasing) still participates in blur, producing dark halos/backgrounds on the transparent layer area. Should be `ignore_alpha 0.5` at minimum, or blur should be disabled for the notification window.

## Fix

**Fix 1** — Add to `Configs/.local/share/hyde/wallbash/theme/swaync.dcol` (and `Configs/.config/swaync/style.css`):

```css
notificationwindow, blankwindow {
    background: transparent;
    background-color: transparent;
}
```

**Fix 2** — In `Configs/.config/hypr/windowrules.conf`, change:
```
layerrule = ignore_alpha 0,match:namespace swaync-notification-window
```
to:
```
layerrule = ignore_alpha 0.5,match:namespace swaync-notification-window
```
(or disable blur for the notification window entirely)

## Environment

- **HyDE commit**: `7493044` (master)
- **Hyprland**: 0.47.x – 0.53.x (reproduced on 0.53.3)
- **swaync**: 0.12.4
- **OS**: Manjaro Linux
- **GPU**: Dual monitor setup (DP-1 2560×1440 + HDMI-A-1 1080×1920 portrait)
- **Relevant HyDE files**:
  - `Configs/.config/hypr/windowrules.conf` (layerrule section at bottom)
  - `Configs/.local/share/hyde/wallbash/theme/swaync.dcol`
  - `Configs/.config/swaync/style.css`

Delcado19:
swaync's layer-shell surface spans the full monitor height (`fit-to-screen: true` in `config.json`), but the generated `theme.css` had no transparent background rule for swaync's GTK window node (`notificationwindow`/`blankwindow`). GTK's own dark default background then painted across that entire area, so every notification showed a black rectangle next to the popup — worst on portrait monitors, where the surface is even taller.

## Fix

Adds the missing rule to `Configs/.local/share/wallbash/theme/swaync.dcol` (the wallbash template that generates `theme.css`):

```css
notificationwindow, blankwindow {
  background: transparent;
  background-color: transparent;
}
```

## Deliberately not included

The issue also suggested raising `ignore_alpha` from `0` to `0.5` on the shared `swaync-notification-window` blur layer rule. That rule is shared with rofi/waybar/selection, and the change is speculative (the reporter themselves framed it as "should be ... at minimum"). The transparent-background fix alone accounts for a full-width black rectangle, not just thin blur edges, so I left the shared alpha threshold untouched rather than adjusting it without being able to verify visually.

## Test plan

- [x] `tests/test_swaync_notification_transparency.sh` (new): asserts the GTK transparency rule is present in `swaync.dcol`
- [x] Revert-and-confirm: reverting the fix makes the new test fail with the expected message; restoring it passes again
- [x] Full suite (`sh tests/run.sh`): 44/44 passing

Fixes #1611.

## Summary by CodeRabbit

* **Bug Fixes**
  * Notification windows now remain transparent, preventing an unwanted dark background from covering the monitor.

* **Tests**
  * Added coverage to verify the notification transparency styling is present and correctly configured.

### DIFF
--- a/Configs/.local/share/wallbash/theme/swaync.dcol
+++ b/Configs/.local/share/wallbash/theme/swaync.dcol
@@ -11,6 +11,15 @@ ${confDir}/swaync/theme.css|${WALLBASH_SCRIPTS}/swaync.sh
 @define-color text-color #FFFFFF;
 @define-color text-color-disabled #070707;
 
+/* swaync's layer-shell surface spans the full monitor height (fit-to-screen
+ * in config.json), so its GTK window node has to be transparent -- without
+ * this, GTK's default dark window background paints across that whole area,
+ * not just the notification popup itself (#1611). */
+notificationwindow, blankwindow {
+  background: transparent;
+  background-color: transparent;
+}
+
 /*
  *
  * control center
--- a/tests/test_swaync_notification_transparency.sh
+++ b/tests/test_swaync_notification_transparency.sh
@@ -1,0 +1,21 @@
+#!/usr/bin/env sh
+# swaync's layer-shell surface spans the full monitor height (fit-to-screen
+# in config.json), so without an explicit transparent background on its GTK
+# window node, GTK's default dark background paints across that whole area
+# -- a black rectangle next to every notification, not just the popup (#1611).
+
+. "$(dirname -- "$0")/lib/common.sh"
+
+dcol="$REPO_ROOT/Configs/.local/share/wallbash/theme/swaync.dcol"
+[ -f "$dcol" ] || {
+    fail "swaync.dcol not found at $dcol"
+    finish
+}
+
+grep -Eq 'notificationwindow,[[:space:]]*blankwindow[[:space:]]*\{' "$dcol" ||
+    fail "swaync.dcol has no rule for the notificationwindow/blankwindow GTK nodes"
+
+awk '/notificationwindow,[ \t]*blankwindow[ \t]*\{/,/\}/' "$dcol" | grep -q 'background:[[:space:]]*transparent' ||
+    fail "the notificationwindow/blankwindow rule does not set a transparent background"
+
+finish


kRHYME7:
can you sahre your swaync config?

Delcado19:
This isn't a personal config -- it's HyDE's own stock shipped default, unmodified: Configs/.config/swaync/config.json has "fit-to-screen": true, "notification-window-width": 400. theme.css is generated from Configs/.local/share/wallbash/theme/swaync.dcol (the file this PR edits) via wallbash on every theme reload -- not hand-edited. So this reproduces on a fresh install with zero customization, no config to share beyond what's already in the repo.

kRHYME7:
can you try to use ` user`. 


from wiki
```
cssPriority
type: string
default: application
values: application, user
description: Which GTK priority to use when loading the default and user
CSS files. Pick "user" to override XDG_CONFIG_HOME/gtk-3.0/gtk.css
```

Delcado19:
Already the case -- HyDE's shipped config.json has "cssPriority": "user" set (line 6). So this isn't the missing piece; it's actually why this fix works: at user priority, theme.css's new background: transparent rule properly overrides the GTK theme's own window background instead of being overridden by it.

kRHYME7:
can you provide a screenshot. just for reference as I can't repro

Delcado19:
Tried reproducing on two different monitors here (1920x1080 laptop panel and a 2560x1440 external, the latter giving a near-full-height 400x1402px notification layer via hyprctl layers -- much closer to the reported conditions). No visible black rectangle on either, even with pixel sampling across the full layer height. So I can't produce a "before" screenshot either -- this looks like it depends on something in the reporter's setup beyond just monitor height (GTK theme, or the portrait orientation specifically). The fix itself (adding the missing background: transparent rule) is still correct per swaync's own default CSS and doesn't regress anything either way.

PR #2079, opened 2026-09-11
fix(theme): accept compatible optional archives

### LINKED ISSUE #2077: [Bug]: Theme patcher rejects valid theme archives
### Description

The theme patcher rejects several public themes even though their repositories contain a valid HyDE theme directory and the referenced archives are present. The validation compares the exact font or cursor value from `hypr.theme` with archive contents and aborts the import when the archive is named differently or an optional resource is absent.

### Affected themes

- **Nightbrew**: `hypr.theme` selects `Bibata-Modern-Ice`, but the repository has no `Cursor_*.tar.*` archive for that cursor.
- **Monterey Frost**: `hypr.theme` selects `SF Pro Rounded` for `FONT`, `DOCUMENT_FONT` and `MONOSPACE_FONT`, while the repository provides `Font_SF Pro Rounded Regular 10.5.tar.xz`; the mandatory checks reject all three values.
- **Electra**: `hypr.theme` selects `Noto Sans`, but the repository contains no matching `Font_*.tar.*` archive.

The repositories are reachable and contain `Configs/.config/hyde/themes/<name>`. Import succeeds for themes whose archive names match the values exactly. The patcher should either accept the archive's actual top-level name, treat missing font/cursor packages as optional when the theme can still be imported, or report a clear actionable error without aborting the whole theme import.

### Reproduction

Run the theme patcher for each repository, for example:

```bash
hyde-shell theme.patch.sh "Monterey Frost" https://github.com/rishav12s/Monterey-Frost
```

Observed errors include:

```text
font set in hypr.theme does not exist in Font_*.tar.*
document-font set in hypr.theme does not exist in Document-Font_*.tar.*
monospace-font set in hypr.theme does not exist in Monospace-Font_*.tar.*
cursor set in hypr.theme does not exist in Cursor_*.tar.*
```

### Expected behavior

Themes with a valid HyDE config should install, with missing optional resources skipped and clearly reported. Archive-name validation should account for the actual archive and directory names supplied by the theme repository.

Delcado19:
## Summary

- accept compatible archive directory names such as `SF Pro Rounded` and `SF Pro Rounded Regular 10.5`
- treat missing non-GTK font, cursor, icon, and other optional packages as warnings so the theme can still be imported
- keep the GTK archive requirement mandatory

## Root cause

The theme patcher required an exact match between the value in `hypr.theme` and the top-level directory inside an archive. It also marked every declared but missing non-GTK package as fatal. Several valid community themes use a family name in `hypr.theme`, a family-plus-style archive directory, or omit optional packages entirely.

## Validation

- `bash -n Configs/.local/lib/hyde/theme.patch.sh`: passed
- isolated Monterey Frost import: archive validation passed; missing optional font variants reported as warnings
- isolated Electra import: missing optional Noto Sans package reported as a warning; import continued
- `git diff --check`: passed

## Risk

Low. GTK remains mandatory. Optional resources are still extracted when present; only validation of absent optional resources changes from an error to a warning.

Closes #2077

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved package archive validation by correctly handling package directory names containing spaces or dashes.
  - Prevented false validation failures caused by surrounding whitespace in package values.
  - Optional package mismatches now generate a note and allow processing to continue, while mandatory mismatches continue to stop processing as required.

### DIFF
--- a/Configs/.local/lib/hyde/theme.patch.sh
+++ b/Configs/.local/lib/hyde/theme.patch.sh
@@ -176,11 +176,34 @@ check_tars() {
         if [[ $gsVal =~ ^\$\{?[A-Za-z_][A-Za-z0-9_]*\}?$ ]]; then
             print_log -warn "Variable $gsVal detected! " "be sure $gsVal is set as a different name or on a different file, skipping check"
         else
+            # Theme authors commonly put a font style/size in the archive
+            # directory name while hypr.theme contains only the family (for
+            # example `SF Pro Rounded` vs `SF Pro Rounded Regular 10.5`).
+            # Compare trimmed names and accept an archive whose top-level
+            # directory starts with the configured value.
+            gsVal="${gsVal##+([[:space:]])}"
+            gsVal="${gsVal%%+([[:space:]])}"
             print_log -g "[pass]  " "hypr.theme :: [$gsLow]" -b " $gsVal"
             trArc="$(echo "$all_tar_files" | grep "/${inVal}_")"
-            [ -f "$trArc" ] && [ "$(echo "$trArc" | wc -l)" -eq 1 ] && trVal="$(basename "$(tar -tf "$trArc" | cut -d '/' -f1 | sort -u)")" && trVal="$(echo "$trVal" | grep -w "$gsVal")"
+            trVal=""
+            if [ -f "$trArc" ] && [ "$(echo "$trArc" | wc -l)" -eq 1 ]; then
+                while IFS= read -r candidate; do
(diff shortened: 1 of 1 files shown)


Delcado19:
Thanks for the review. Fixed in `b251f48e` by enabling Bash `extglob` before `check_tars` is defined and called, so the existing leading/trailing whitespace trimming works as intended.

Validation: `bash -n Configs/.local/lib/hyde/theme.patch.sh` and `git diff --check` both pass.

PR #2080, opened 2026-09-11
fix(wallpaper): handle missing generated thumbnails

### LINKED ISSUE #2078: [Bug]: Wallpaper cache moves missing quad thumbnail
### Description

The wallpaper cache prints an error while importing or switching themes when ImageMagick does not create the temporary quad thumbnail file:

```text
mv: cannot stat '/home/delcado/.cache/hyde/thumbs/0392149aa9ca0c8b4f0a077aedc4eff0d113216b.quad.png': No such file or directory
```

### Cause

`wallpaper/cache.sh` runs `mv "$thmbDir/$x_hash.quad.png" "$thmbDir/$x_hash.quad"` after ImageMagick returns success, without checking that the temporary output file exists. On the affected wallpaper the generated file is absent, so the cache reports a raw `mv` error.

### Expected behavior

The cache should verify the temporary file before moving it and emit a contextual warning (or skip that thumbnail) instead of exposing a raw `mv` error during theme installation. A failed optional thumbnail should not make theme installation look broken.

Delcado19:
## What changed

The wallpaper cache now checks that ImageMagick actually created temporary square and quad thumbnails before moving them into the cache. Missing optional thumbnail output is skipped with a contextual warning instead of exposing a raw `mv: cannot stat` error.

The same guard is used by normal and forced cache generation.

## Root cause

The cache chained `mv` directly after ImageMagick. When ImageMagick returned without creating the requested output file, `mv` reported the missing path directly.

## Validation

- `bash -n Configs/.local/lib/hyde/wallpaper/cache.sh`
- Focused Bash test covering successful moves and missing-file warnings
- `git diff --check`

Risk is limited to thumbnail cache generation; existing generated files and successful moves are unchanged.

Closes #2078

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved wallpaper thumbnail caching when multiple cache workers run concurrently.
  - Prevented duplicate thumbnail moves from failing when the target thumbnail already exists.
  - Added warnings for other thumbnail move failures while preserving an error status.
  - Retained safe handling for guarded thumbnail generation in standard and forced cache generation.

### DIFF
--- a/Configs/.local/lib/hyde/wallpaper/cache.sh
+++ b/Configs/.local/lib/hyde/wallpaper/cache.sh
@@ -22,7 +22,15 @@ move_generated_thumbnail() {
     local source="$1"
     local target="$2"
     if [ -f "$source" ]; then
-        mv -- "$source" "$target"
+        if mv -- "$source" "$target"; then
+            return 0
+        elif [ -e "$target" ]; then
+            # Another cache worker may have created the target concurrently.
+            return 0
+        else
+            printf 'Warning: failed to move generated wallpaper thumbnail "%s" to "%s"\n' "$source" "$target" >&2
+            return 1
+        fi
     else
         printf 'Warning: generated wallpaper thumbnail "%s" is missing; skipping "%s"\n' "$source" "$target" >&2
     fi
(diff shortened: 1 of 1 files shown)


Delcado19:
Addressed in `a58eaeda`.

`move_generated_thumbnail` now treats a failed `mv` as successful when the destination already exists, covering concurrent cache workers. Other move failures emit a contextual warning and return failure. The existing source-file check is preserved, and both source/target variables are used directly.

Validation passed:
- `bash -n Configs/.local/lib/hyde/wallpaper/cache.sh`
- focused Bash tests for successful moves, missing sources, concurrent targets, and real move failures
- `git diff --check`

PR #2086, opened 2026-09-12
feat(gpuinfo): use LACT for cross-vendor metrics

Delcado19:
## Summary
- read GPU identity and live metrics through the LACT daemon socket
- keep the six-field Waybar schema stable and render unavailable values as `N/A`
- add interactive and silent installer controls for installing `lact` and enabling `lactd`

## Validation
- all `tests/test_gpuinfo_lua_*.sh` checks pass
- `bash -n Scripts/install.sh` passes
- live LACT query verified against an NVIDIA GeForce RTX 2070

The adapter does not substitute CPU, generic sensor, or battery values when LACT cannot provide a GPU metric.

## Summary by CodeRabbit

- **New Features**
  - Reworked GPU monitoring with unified vendor detection, temperature, utilization, power, clock, and status information.
  - Added optional LACT support for normalized GPU metrics and monitoring.
  - Added NVIDIA, AMD, and Intel GPU selection, toggling, resetting, and suspend-aware monitoring.

- **Bug Fixes**
  - Improved Waybar GPU module detection and vendor selection.
  - Added safer handling for missing, invalid, or unavailable sensor and GPU data.

- **Tests**
  - Expanded automated coverage for GPU detection, monitoring, output, state, and error handling.


kRHYME7:
If we are using lact then might as will keep the gpuinfo.sh script and add a explicit lact.lua wrapper. A separate waybar module etc. This way we can do some test with working hyde original script. 

I tried  lact last year (or last last) and it's not yet good that time. I only focus on nvidia gain that time and HyDE's principles to be mindful with dependencies.

Lactd is a daemon and also a dependency. if you want to take this route which is better, then leave old scripts and build a separate lact script and modules for waybar. Then we might eventually swapped the module declared in waybar layouts. That will make it a smoother transition.

Delcado19:
Superseded by #2096 — a fresh implementation from plain `dev` (no `gpuinfo.lua`), keeping `gpuinfo.sh`/`gpuinfo.lua` completely untouched per the review discussion here.

PR #2091, opened 2026-09-13
fix(install): extend locale adaptation to SDDM and hyprlock clocks

Delcado19:
## Summary
- #2042 adapted the waybar clock's 12h/24h format and date order to the system locale, but left two other clocks with the same hardcoded-English-12h problem: the SDDM greeter theme (Candy/Corners, and any `Sddm_*` theme bundled with a color theme) hardcodes `HourFormat`/`DateFormat` in its `theme.conf`, and every hyprlock preset hardcodes a 12h `date` format string.
- `restore_locale.sh` now switches hyprlock's preset clocks to 24h alongside the waybar clock, reusing the already-computed 12h/24h decision. SDDM's `HourFormat`/`DateFormat` are blanked once, at theme-extraction time in `install_pst.sh` and `theme.patch.sh`, so Qt's own locale-aware `Locale.ShortFormat`/`LongFormat` takes over instead.
- Also fixes a second, pre-existing bug found while re-verifying #2042: the waybar date-order fix matched only the shipped default's literal `%d·%m·%y` segment, so a single run under a different locale would permanently wedge the date order, with no way for a later run (even under the correct locale) to self-correct. Now matches any `%[dmy]·%[dmy]·%[dmy]` segment instead of one fixed literal.

## Test plan
- [x] New `tests/test_restore_locale_format.sh` covering both sed transforms and the order-agnostic date resub, run via `tests/run.sh` (full suite green apart from the pre-existing, unrelated `test_gpuinfo` failure)
- [x] Sandboxed dry-run and real-run against copies of the real hyprlock presets under both a 24h (`de_DE`) and 12h (`en_US`) locale, confirming the 24h locale converts every preset and the 12h locale leaves them untouched, and that a second run is a no-op
- [x] Applied to a live install: hyprlock screenshot-verified (24h time, no AM/PM, correctly localized weekday) and SDDM greeter visually confirmed correct after a reboot

## Summary by CodeRabbit

* **New Features**
  * Locale-aware time and date formatting now applies to SDDM greeters and hyprlock lockscreen presets.
  * 12-hour clock presets automatically switch to 24-hour formatting when required by the system locale.
  * Date formatting adapts across different day-month-year orders.

* **Bug Fixes**
  * SDDM themes now follow system locale settings without altering unrelated configuration values.

* **Tests**
  * Added coverage for locale restoration across hyprlock, SDDM, and waybar.


kRHYME7:
can we do multi file for sddm ?   I am not comfortable  with sed in this situation.

kRHYME7:
https://github.com/HyDE-Project/sddm-themes

example we can use deez-dots for sddm install the theme and after that use deez too to patch a theme like we keep 2 file if sddm .conf file do not permit multi config file.

Delcado19:
Fair, `sed -i` on the extracted vendor file wasn't a great look. Rather than mutating it in place, I'll ship a HyDE-maintained `theme.conf` per bundled theme (Candy/Corners) with `HourFormat`/`DateFormat` already blank, and have `install_pst.sh`/`theme.patch.sh` copy that in after extraction instead of sed-patching the vendor one — same effect, but the intended file content is fully visible in the PR diff instead of a transformation.

Fully moving SDDM theme delivery onto deez/`sddm-themes` (so it's a normal dot like everything else) sounds right longer-term, but that's a bigger migration than this locale fix — happy to scope that separately if you want to go there.

Delcado19:
@kRHYME7 Update on the removed \`restore_locale.sh\` call: turns out we don't need it back for the clock/date part at all.

Waybar's \`clock\` module (unlike \`simpleclock\`) renders in the plain C locale unless you prepend \`L\` to the format string — \`{:L%X}\` renders the system's own locale-preferred time, live, on every render, no install-time patching needed. Same story for the hyprlock presets' hardcoded \`date +"%I:%M %p"\` — swapping in \`%X\` makes \`date\` itself locale-aware.

Opened #2112 with both fixes (Waybar clock + 5 hyprlock presets), verified against every file's exact command line under a real locale.

That leaves \`restore_locale.sh\`'s keyboard-layout/keybind-remap logic as the only part that would still need the call restored — happy to open a separate PR for just that piece if you're fine with it, or if you remember what originally broke when it was enabled, that'd help us not repeat it.

kRHYME7:
okk

PR #2106, opened 2026-09-16
fix(install): let swaync_config own its files instead of conflicting

### LINKED ISSUE #2103: [Bug]: extra.toml deploy fails — swaync_config dot conflicts with swaync and waybar dots
### Description
`./install.sh -r` (restore, dev branch) fails with "[DEEZ-DOTS] ERROR :: Extra dotfiles deployed with failures". The `swaync_config` dot added in commit `26602069` ("fix(notifications): install only Dunst by default") can never deploy: its file entries conflict with files already owned by the `swaync` and `waybar` dots.

`Scripts/dots/swaync-config.toml` declares two file entries:
1. `Configs/.config/swaync` → `${XDG_CONFIG_HOME}/swaync` — also owned by the legacy `swaync` dot (`Scripts/dots/swaync.toml`, still present in the tree, no longer in any group).
2. `Configs/.local/share/waybar` → `${XDG_DATA_HOME}/waybar` (`menus/swaync.xml`) — already owned by the `waybar` dot, which syncs the entire `Configs/.local/share/waybar` tree including `menus/`.

So the extra deploy always emits `File conflict: .../waybar/menus/swaync.xml (owned by waybar)` on every machine (fresh installs included), plus `File conflict: .../.config/swaync/* (owned by swaync)` on machines upgraded from a pre-refactor install. This matches deez-dots' `_check_conflicts_from_bundle` ownership-index check.

Tested on a checkout containing commit `26602069` (also present in `origin/dev` as of now). Reproduced independently of a full reinstall:

```bash
~/.local/state/hyde/python_env/bin/deez \
  --source /home/spacious/HyDE \
  --config /home/spacious/HyDE/Scripts/dots-groups/extra.toml \
  dots --skip-git --deploy
```
```
[err] File conflict: /home/spacious/.config/swaync/config.json (owned by swaync)
[err] File conflict: /home/spacious/.config/swaync/style.css (owned by swaync)
[err] File conflict: /home/spacious/.config/swaync/user-style.css (owned by swaync)
[err] File conflict: /home/spacious/.local/share/waybar/menus/swaync.xml (owned by waybar)
[err] Deploy failed for 'swaync_config': exited with status 1
```

### Steps to Reproduce
1. Check out `dev` (contains `26602069`).
2. Run `./install.sh -r` (or `deez dots --deploy <extra config>`).
3. Observe: core + fish dots deploy, extra group fails on `swaync_config`.
4. End of run: "Some dots were not deployed. Deal with the failures reported above and run the restore again."

### Expected Behavior
`swaync_config` deploys its config without conflicting; restore completes with "Dotfiles deployed".

### Actual Behavior
Extra deploy always fails for `swaync_config` due to file conflicts with the `swaync` and `waybar` dots, so the restore always reports a failure on this branch.

### Screenshots
N/A

### Environment
- Display Resolution: 2560x1440
- Graphics Driver: amd
- Display Server: Wayland
- Branch: `dev`
- `deez-dots` 0.2.0 (HyDE python env), swaync 0.12.6-1 installed but not running (Dunst default)

### Additional Information
Suggested fix (verified to resolve it): in `Scripts/dots/swaync-config.toml`, drop the second file entry — the waybar menu lives under `~/.local/share/waybar/menus` and is already owned by the `waybar` dot (`Configs/.local/share/waybar` → `${XDG_DATA_HOME}/waybar`). The `.config/swaync` conflict additionally requires releasing the legacy `swaync` dot's ownership (e.g. removing its stale deez manifest on existing installs).

Remaining design tension to address: opting into the SwayNC daemon via `deez dots --deploy swaync` will still conflict with `swaync_config` on `.config/swaync`, since both dots own those paths. Consider making `swaync.toml` deploy only the daemon dependency, or route `.config/swaync` through a single owner.

Note on prior migrations: `#26602069` moved `swaync.toml` out of `extra.toml` and `notification-daemon.toml`, but `swaync.toml` and its manifest ownership were left in place on upgraded systems.

Delcado19:
## Summary
Fixes #2103. #2084 (26602069) split the old `swaync` dot into an opt-in daemon install and a `swaync_config` dot for the config files, but two of its file entries were never carved out of the dots that already owned those paths, so `swaync_config`'s deploy always fails with a file conflict — on every install, fresh or upgraded:

- `Configs/.config/swaync` is also declared by the now-unreferenced legacy `swaync` dot. A machine that had it deployed before the split keeps a manifest recording it as the owner forever, since nothing deploys `swaync` again to update or clear that record.
- `Configs/.local/share/waybar/menus/swaync.xml` sits inside the `menus` directory the `waybar` dot syncs wholesale, so waybar's manifest claims it on every install, fresh or not.

**Update:** a real user ([@JunaidQrysh](https://github.com/JunaidQrysh)) tested the first version of this PR and hit the exact same conflict — the fix as originally written didn't actually work on an already-installed machine. Root cause: deez-dots reuses a dot's cached bundle whenever the dot's own *source* files haven't changed since it was built (mtime-based). The `ignored_paths` change below is metadata on the dot definition, not a source file, so it never invalidated `waybar`'s existing cached bundle — `waybar` kept reusing the old bundle that still claims `menus/swaync.xml`, so `swaync_config` kept failing the same way on every restore, forever. Fixed by also clearing the stale cached bundle (not just the manifest) and by running that cleanup *before* the dots deploy step instead of only at the end of `install.sh -r`, where migrations normally run (too late to help the same run).

Changes:
- Delete the dead `swaync` dot (`Scripts/dots/swaync.toml`) — no group has included it since #2084, it only still existed to keep conflicting.
- Exclude `menus/swaync.xml` from the `waybar` dot's directory sync via `ignored_paths`, so `swaync_config` is the sole owner going forward.
- Add a migration (`Scripts/migrations/v26.9.2.sh`) that clears the legacy `swaync` and `waybar` dots' stale manifests *and* their cached build bundles, without touching the actual config files on disk (`swaync_config`'s own sync re-lays them right after).
- Call `run_pending_migrations` a second time in `install.sh`, right before the dots deploy step, so this migration's cleanup takes effect in the same restore instead of one run too late. Safe to call twice — applied migrations are tracked in a state file, so the existing end-of-restore call just finds nothing new pending.

## Test plan
- [x] Reproduced the exact conflict from #2103 on the pre-fix tree using its repro command: `deez --source $(pwd) --config Scripts/dots-groups/extra.toml dots --skip-git --deploy` — same four `File conflict` errors.
- [x] Reproduced @JunaidQrysh's follow-up failure by manually reconstructing a pre-fix `waybar` manifest + cached bundle tarball on top of the first fix commit — confirmed it still fails the same way, proving the first version was insufficient.
- [x] Ran `core.toml` then `extra.toml` deploy after the corrected migration (matching `install.sh`'s real order): `waybar` rebuilds fresh without `menus/swaync.xml`, `swaync_config` deploys with 0 conflicts — in one pass, no `--rebuild` needed.
- [x] Verified the migration is idempotent via the real `run_pending_migrations` state tracking: applies once, is correctly skipped on a simulated second (end-of-restore) call.
- [x] Restarted Waybar after the change — starts cleanly, no new errors.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed dotfile deployment failures caused by conflicting SwayNC and Waybar configuration ownership.
  - Prevented stale deployment records and cached bundles from repeatedly triggering file-conflict errors.
  - Ensured existing configuration files remain intact during cleanup and migration.

- **Maintenance**
  - Improved cleanup of outdated deployment manifests and cached bundles.
  - Cleanup now continues across applicable items and reports any failures after processing.
  - Deployment migrations now run before restoring dotfiles to apply ownership changes reliably.


JunaidQrysh:
Does not work,
```
[err] File conflict: /home/qrysh/.local/share/waybar/menus/swaync.xml (owned by waybar)
[err] Deploy failed for 'swaync_config': exited with status 1
ok
[err] Deploy finished with 1 failed dot(s):
[err]   swaync_config: exited with status 1
ERROR :: [DEEZ-DOTS]   ERROR  :: Extra dotfiles deployed with failures
```

Delcado19:
@JunaidQrysh Thanks for testing — you were right, the previous commit didn't actually fix it.

Root cause: deez-dots reuses a dot's cached bundle whenever the dot's own source files haven't changed since it was built. The `ignored_paths` change to `waybar.toml` is metadata on the dot definition, not a source file, so it never invalidated `waybar`'s existing cached bundle on an already-installed machine — it kept reusing the old bundle that still claims `menus/swaync.xml`, so `swaync_config` kept failing exactly the same way on every restore.

Pushed a new commit that also clears the stale cached bundle (not just the manifest), and fixes *when* that cleanup runs: migrations normally run once at the very end of `install.sh -r`, after both `core.toml` and `extra.toml` have already deployed and already failed for that run. There's now a second, identical migrations call right before the deploy step, so the stale cache gets cleared in time for `waybar` to rebuild correctly and `swaync_config` to deploy clean in the same run.

Reproduced your exact failure here with a manually reconstructed pre-fix manifest + cached bundle, then verified the fix resolves it end-to-end (one pass, no `--rebuild` needed). Would you be able to test `fix/swaync-config-dot-conflict` again?

Scripts/install.sh
kRHYME7: why doing this for a single fix? 

how about version bump hyde and swaync dots so it triggers an uninstall and install in deez-dots?

JunaidQrysh:
fixed, but two issues-
1.if i install swaync and uninstall dunst i get 'Executable not found: dunst' everytime i login even though swaync works perfectly
2. waybar notification module does work, from journal -```[warning] There is no configuration for 'custom/notifications', element will be hidden```

kRHYME7:
> 1.if i install swaync and uninstall dunst i get 'Executable not found: dunst' everytime i login even though swaync works perfectly
 
 
 you should  declare that in your config.toml. HyDE expects dunst and no fallbacks.  (your system will try to use any available one though). HyDE does not restrict users to use other daemons but it requires them to at least declare it.

https://github.com/HyDE-Project/HyDE/blob/master/Configs/.local/share/hyde/schema/config.md#desktopstart
 
 ```
 
[desktop.start]
notifications = "swaync"

```

Delcado19:
@kRHYME7 You're right — traced the load order: \`variables.lua\` sets the \`dunst\` default, \`dynamic.lua\` merges \`config.toml\` over it, \`start_up.lua\` executes the result last. The \`[desktop.start]\` override already handles this cleanly and consistently for every entry. The auto-detect fallback I shipped in #2109 was an unnecessary special case that duplicates that mechanism. Reverting it back to the plain \`dunst\` default in a follow-up PR. Thanks for the catch.

PR #2113, opened 2026-09-16
fix(shaders): preview coalescing, damage-tracking consent, rofi menu overflow

Delcado19:
## Description

Three related fixes to the shader picker:

- `fix(shaders): preserve state and require consent for damage tracking` — the picker now asks for confirmation before enabling damage tracking for shaders that need it, and correctly restores the prior state when the menu is cancelled.
- `fix(shaders): coalesce rapid and repeated preview requests` — rapid selection changes in the picker (arrow-key scrubbing) no longer fire overlapping preview/apply calls.
- `fix(rofi): keep cursor-follow menus from overflowing off-screen` — `get_rofi_pos()` chose the anchor edge purely from which half of the monitor the cursor was in, without checking whether the (fixed-height) rofi window actually fit in the remaining space in that direction. A menu opened near the middle of the screen could grow past the far edge and get clipped. This surfaced on the shaders picker, but the positioning helper (`luautils/rofi/pos.lua`) is shared by every rofi menu that follows the cursor (wallpaper, theme, clipboard, ...), so the fix is in the shared function, not the shader script.

## Type of change

## Checklist

## Additional context

The rofi positioning fix includes a small self-check (`lua5.1 Configs/.local/lib/hyde/luautils/rofi/pos.lua`) covering the overflow-clamp logic on both axes, in addition to the existing `.github/scripts/test_shaders.lua` coverage for the shader picker itself.

## Summary by CodeRabbit

- **New Features**
  - Shader selection now supports safer previews, explicit confirmation for high-GPU damage-tracking changes, reliable persistence, and recovery from interrupted or failed operations.
  - Shader metadata validation provides clearer handling for missing files, unsupported hooks, and unsafe display names.
  - Cursor-following menus better fit within screen boundaries, including oversized windows and custom minimum sizes.

- **Bug Fixes**
  - Improved shader state rollback and restoration when applying or switching shaders.
  - Fixed menu positioning near screen edges and in constrained spaces.

- **Documentation**
  - Added guidance about GPU impact and setting restoration when disabling damage tracking.

- **Tests**
  - Added automated coverage for shader workflows, recovery, consent handling, caching, and menu positioning.


kRHYME7:
please test suite is in github.com/Hyde-project/tests

kRHYME7:
@Delcado19  can you move the test to the test repo  ?

Delcado19:
Pushed a follow-up commit addressing both CodeRabbit findings from today's review:
- shaders.lua: the crash-recovery menu file is now only removed once the restore actually succeeds, so a rejected/crashed restore doesn't lose the pre-menu snapshot.
- rofi.lua: the follow-cursor overflow estimate now reads the resolved theme's own window size instead of always assuming the clipboard dropdown's 23em x 30em.

Added regression coverage for both in the tests submodule (HyDE-Project/tests#4); the submodule pointer here will follow once that's merged. Also dropped .github/scripts/test_shaders.lua and its dedicated CI step, now redundant with the submodule's own shader test.

PR #2119, opened 2026-09-19
fix(theme): resolve theme-repo branches via git instead of the GitHub API

Delcado19:
## Summary
- Fixes one of the root causes identified in #2118: `theme.patch.sh` resolved a bare theme-repo URL's branch via an unauthenticated GitHub REST API call (`curl .../branches | jq`). That endpoint is capped at 60 requests/hour for anonymous callers, and the call had no error handling — a rate-limited or malformed response left the branch array empty, so `branch` silently became `""` and `git clone -b ""` failed with an opaque message. Bulk theme imports could hit that cap mid-session, so only the themes processed after the limit was hit failed (matches the reporter's Obsidian Purple / Monterey Frost symptoms).
- Replaces it with `git ls-remote --heads`, which uses the git protocol rather than the GitHub REST API and isn't subject to that rate limit — no token, no setup, works the same for a single branch or a multi-branch repo (still prompts to select, as before).
- Adds explicit error handling: an unreachable repo or a repo with no branches now exits with a clear `[ERROR]`, instead of silently continuing with an empty branch name.

## Test plan
- [x] New regression test in HyDE-Project/tests#6, covering single-branch auto-resolve, multi-branch select, an unreachable repo failing loudly, and that `curl` is never invoked
- [x] Verified the test fails against the pre-fix script (4/4 cases) and passes against the fix
- [x] Ran `tests/run.sh` locally: no new regressions beyond pre-existing, unrelated failures
- [x] Manually exercised `theme.patch.sh` against local single- and multi-branch git fixtures and an unreachable repo URL

Related: HyDE-Project/tests#6 (regression test, separate repo/PR per the tests submodule split)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved branch detection for Git repositories.
  - Avoided GitHub API rate limits during branch selection.
  - Added clear errors when a repository cannot be reached or has no branches.
  - Fixed handling for repositories with exactly one available branch.
  - Improved reliability when listing available branches.

- **Tests**
  - Updated the test suite reference to include the latest test changes.


rubiin (changes requested):
tests fail

Delcado19:
@rubiin the 4 failures are pre-existing and unrelated to this PR's change (theme.patch.sh branch resolution) — they'd show up on any PR that pulls in the `tests` submodule at its current state:

- `test_lact_wrapper`: tests a module (`lact.lua`) that only exists on the still-open HyDE-Project/HyDE#2096, not yet on `dev`. Already tracked in HyDE-Project/tests#5.
- `test_rofi_pos` / `test_rofi_selector`: test the overflow-aware positioning rewrite that lives in the still-open HyDE-Project/HyDE#2113, also not yet on `dev`.
- `test_shaders`: two issues bundled — missing `lfs` (LuaFileSystem) on the CI runner (fixed in HyDE-Project/tests#7) and, once that's fixed, the same #2113-dependency as above (shader damage-tracking consent).

All three feature-dependent ones should go green once #2113 merges. My new regression test (`test_theme_patch_branch_resolve`) passes cleanly. Happy to open a tracking issue for the #2113 dependency if that's useful, or point me at anything you'd like changed on this PR specifically.

Delcado19:
Update: rebased onto `dev` now that #2113 merged. As expected, `test_rofi_pos`, `test_rofi_selector`, and `test_shaders` are all green now. The only remaining CI failure is `test_lact_wrapper`, which is unrelated to this PR — it's blocked on the still-open HyDE-Project/HyDE#2096 and already tracked in HyDE-Project/tests#5.

kRHYME7:
For PR's in the future is it possible for a  test to point on its own branch right? 

Hyde - pr branch - you pull latest changes and test in you local should be pointing to the pr tests branch 
tests  - pr branch  - you pull any change

and should provide greenlight? I will be mindful soon on merging tests while a pr was still being work on.

kRHYME7:
nitpicking: Can we color errors so we can see the block. Hehe thank you! 

image

Delcado19:
That's exactly the pattern I've been using here: point this PR's `tests` submodule pointer at the tests PR's own branch while both are still open (so this PR's CI exercises it), then re-point to the merged commit once the tests PR actually lands. Happy to keep doing that going forward if it's useful as a general convention.

On the colour nitpick — already done: HyDE-Project/tests#9. Colours `ok`/`skip`/`fail`/`FAILED` (green/yellow/red), only when attached to a terminal or running in CI, plain otherwise so redirected output stays exact. This PR's submodule pointer now includes it too.

kRHYME7:
is this pr ready to be merge ?

Delcado19:
The theme.patch.sh fix itself is ready — code's done, CI is green.

One loose end though: the `tests` submodule pointer right now points at a temporary integration branch in my fork (combining the still-open tests#8 and tests#9), not a real upstream commit. I wouldn't want that landing in `dev` as-is.

I'd suggest holding off until tests#8 and tests#9 both merge, then I'll do one final pointer bump to the real HyDE-Project/tests main commit — after that it's genuinely ready. Should be quick once those land.

Delcado19:
Both tests#8 and tests#9 merged, so I did the final pointer bump to the canonical `HyDE-Project/tests` main commit. CI is green across the board now — this is ready to merge.

rubiin:
btw why did we move tests to another repo. Makes adding the tests quite hectic considering we have to merge the pr there

PR #2120, opened 2026-09-19
fix(waybar): make Caffeine mode survive a config hot-reload

### LINKED ISSUE #2117: [Bug]: Caffeine mode silently deactivates on any Waybar reload (theme/wallpaper change), not just --hide
Related to #1177, closed as completed after `cbff6d27` ("fix: improve performance on theme change") replaced Waybar's kill+respawn with a SIGUSR2 hot-reload. That fixed the "hiding bug" but not the actual root cause: SIGUSR2 still re-parses the whole config and reinitializes every module, including `idle_inhibitor`. Waybar keeps that module's toggle state only in process memory — nothing restores it after a reload. So Caffeine Mode silently resets to "deactivated" on **any** `waybar.py` call that ends in `restart_waybar()`, not just `--hide` — routine theme/wallpaper switches trigger it too, with zero notification to the user.

**Steps to reproduce:**
1. Enable Caffeine mode via the waybar `idle_inhibitor` module.
2. Trigger anything that calls into `waybar.py`'s default path (e.g. wallpaper/theme change).
3. Check `journalctl` — waybar hot-reloads via SIGUSR2 (same PID, not a restart) and reinitializes `idle_inhibitor`; hypridle's listeners fire again on the normal timeouts as if Caffeine was never on.

**Live reproduction (2026-09-18, master @ 1a1ed2e2):**
```
05:58:00  waybar/hypridle start, Caffeine enabled manually shortly after
08:38:17  waybar[1984] "Including resource file: .../idle_inhibitor.jsonc"
          (SIGUSR2 hot-reload triggered by wallpaper.sh/theme reload — same PID, not a restart)
08:41:18  hypridle "Idled: rule ... / Running brightnessctl -s && brightnessctl s 1%"
          (60s dim listener fires — Caffeine no longer inhibiting)
```

**Expected:** Caffeine Mode survives a Waybar config hot-reload, or the user gets a visible signal when it resets.
**Actual:** Silently turns off, screen dims/locks/DPMS-off on the normal schedule.

**Possible direction:** Waybar's native `idle_inhibitor` has no state-restore hook across reloads, so this may need to move off the built-in module — e.g. a `custom` module backed by an external process (systemd-inhibit or a small standalone wlr-idle-inhibit client) with its own state file, decoupled from Waybar's in-process module state. Happy to help test — wanted to flag this with concrete evidence first since #1177 looks fixed on paper but isn't in practice.

Delcado19:
## Summary
Fixes #2117: Waybar's native `idle_inhibitor` module type keeps its activated/deactivated state only in Waybar's own process memory. Waybar hot-reloads its config via SIGUSR2 on any routine theme/wallpaper change (not just `--hide`), which reinitializes every module — silently resetting Caffeine mode to off, with zero notification to the user. hypridle then follows its normal idle timeouts as if Caffeine was never enabled.

- Replaces the native `idle_inhibitor` module with a `custom/caffeine` module (`custom-caffeine.jsonc`) backed by a new `caffeine.sh`.
- `caffeine.sh` tracks on/off state in a runtime state file (`$XDG_RUNTIME_DIR`, since this is inherently session/boot-scoped, not a persistent preference) and does the actual inhibiting via a `systemd-inhibit --what=idle:sleep` background process. That process is independent of Waybar's own lifecycle, so a hot-reload can't touch it.
- `hypridle.conf` already has `ignore_systemd_inhibit = false`, so it respects this natively — no custom Wayland idle-inhibit protocol client needed (per the design discussion on #2117).
- Self-heals a stale state file: a dead recorded pid, or one reused by an unrelated process after a reboot, is treated as "off" and corrected rather than trusted blindly.
- Fails loudly (not silently) if `systemd-inhibit` isn't installed, instead of claiming to be activated when it isn't.
- Uses `flock` around the state read-modify-write so a rapid double-click can't corrupt the state file or double-spawn/double-kill the inhibitor.
- Replaced `"idle_inhibitor"` with `"custom/caffeine"` in every shipped layout that referenced it (19 files) — no styles reference `#idle_inhibitor` by name, so no CSS changes needed.

## Test plan
- [x] New regression test in HyDE-Project/tests#10 covering toggle on/off, a dead/reused pid, corrupted state file shapes, a missing `systemd-inhibit` binary, invalid `--sigproc` input, and a concurrent double-toggle race
- [x] Verified the recorded pid is confirmed as a genuine, live `systemd-inhibit` process (not just any live pid) — and that killing it cleans up its child process too (checked against the real `systemd-inhibit` binary directly)
- [x] All touched `.jsonc` layouts validated as syntactically correct
- [x] Full test suite (`tests/run.sh`): 39 cases, 0 failed
- [x] Live end-to-end check on a real Hyprland session, with an isolated throwaway Waybar instance so the actual desktop bar was never touched:
  - Toggled on: confirmed a real `systemd-inhibit` lock via `systemd-inhibit --list`, and confirmed the icon visually renders and changes (outlined cup ↔ filled cup) on click, with a proper Nerd Font set
  - Sent that instance SIGUSR2 (the same signal a theme/wallpaper change triggers) and confirmed the inhibitor lock, state file, and re-queried status all still reported "activated" afterward — the actual bug this PR fixes
  - The test instance then crashed on an unrelated GLib/D-Bus GApplication singleton conflict from running two simultaneous Waybar instances (mine + the real one) — reproduced identically with a totally unrelated plain "clock"-only config, confirming it's a pre-existing multi-instance limitation, not something this change causes; a normal single-instance setup never hits it
  - Cleaned up fully afterward: inhibitor released, test processes killed, temp files removed, real desktop bar unaffected throughout

Related: HyDE-Project/tests#10 (regression test, separate repo/PR per the tests submodule convention)

## Summary by CodeRabbit

* **New Features**
  * Added a custom Waybar caffeine control with status updates, tooltips, and click-to-toggle behavior.
  * Updated supported Waybar layouts to display the caffeine control.

* **Bug Fixes**
  * Improved caffeine state handling by requiring a valid runtime directory for state and lock files.
  * Installation now reports theme restoration failures instead of continuing silently.

* **Tests**
  * Updated the referenced tests revision.


Delcado19:
Good catch on the `XDG_RUNTIME_DIR` fallback — fixed. Worth noting: `hyde-shell`/`globalcontrol.sh` sets its own fallback for `XDG_RUNTIME_DIR` (`/run/user/$(id -u)`), and since caffeine.sh sourced it *before* checking, my first attempt at this fix was actually unreachable dead code. Reordered so the check runs on the raw, un-defaulted value before that source happens (caffeine.sh doesn't use anything else from it). Added a regression test in tests#10 covering it: clear error, non-zero exit, no `/tmp/hyde` created.

The `tests` gitlink pointing at an unmerged commit is expected — that's the tests submodule convention this project uses (point at the tests PR's branch while both are open, re-point to the merged commit once it lands). Will do that follow-up bump once tests#10 merges.

Configs/.local/share/waybar/modules/idle_inhibitor.jsonc
kRHYME7, line 1: Please don't nuke this. Both have different approaches Maybe we can keep it. Also there are downstream layouts that break if you remove this. (like mine)

kRHYME7 (changes requested):

Configs/.local/share/waybar/modules/idle_inhibitor.jsonc
Delcado19: Good catch, thanks. Restored idle_inhibitor.jsonc as-is in 807d43e4 rather than deleting it -- HyDE's own layouts now use custom/caffeine for the hot-reload fix, but the native module definition stays in place for layouts (like yours) that still reference it directly. The two use independent mechanisms (Wayland idle-inhibit protocol vs. systemd-inhibit) and don't conflict if both happen to be present.

Configs/.local/lib/hyde/caffeine.sh
```diff
@@ -0,0 +1,239 @@
...
+# they own, or a symlink) and read or interfere with this user's caffeine
+# state and inhibitor pid (CWE-377). XDG_RUNTIME_DIR is always set on a
+# normal systemd/logind-managed session, so its absence means a broken or
+# unusual environment -- fail clearly instead of degrading into that risk.
+if [ -z "$XDG_RUNTIME_DIR" ]; then
+    echo "Error: XDG_RUNTIME_DIR is not set, refusing to use a shared fallback location" >&2
+    exit 1
+fi
+
+if ! source "$(which hyde-shell)"; then
```
kRHYME7, line 29: safer to use command -v or type -p

rubiin (approved):

kRHYME7 (changes requested):

PR #2121, opened 2026-09-19
fix(wallpaper): route the KIO service menu action through hyde-shell

Delcado19:
## Summary
Related to #2118: the reporter got "Could not find the program 'wallpaper.kon.sh'" clicking "Refresh List" under a file manager's right-click "Set As Wallpaper..." submenu.

Root cause: `hydewallpaper.desktop`'s `Exec=` line calls `wallpaper.kon.sh` by its bare name. That script genuinely exists at `Configs/.local/lib/hyde/wallpaper.kon.sh` and still works — it isn't missing or removed. The actual problem is PATH: `$HOME/.local/lib/hyde` is only added to `PATH` by the interactive shell's `.zshrc`. HyDE's own shipped session-wide environment (`Configs/.config/uwsm/env.d/00-hyde.sh`) only prepends `$HOME/.local/bin`. A GUI app the session launches (a file manager running a KIO service menu action) inherits the session's PATH, not a login shell's — so the bare script name resolves fine from a terminal but fails from the GUI.

Fix: route through `hyde-shell wallpaper.kon.sh` instead of the bare script name. `hyde-shell` itself lives in the always-present `$HOME/.local/bin`, and dispatches to `lib/hyde` scripts by name regardless of the caller's own PATH.

## Test plan
- [x] New regression test in HyDE-Project/tests#11: a static check that no `.desktop` `Exec=` line names a `Configs/.local/lib/hyde/*.sh` script by its bare, unrouted name. Fails against the pre-fix file, passes after.
- [x] Manually verified: `hyde-shell wallpaper.kon.sh` resolves and runs cleanly under a simulated GUI-session PATH (`$HOME/.local/bin:/usr/bin:/bin`, no `$HOME/.local/lib/hyde`), where the bare `wallpaper.kon.sh` name does not.
- [x] Full test suite (`tests/run.sh`): all cases pass

Related: HyDE-Project/tests#11 (regression test, separate repo/PR per the tests submodule convention)

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the wallpaper refresh menu action to run through the shell environment, improving compatibility when refreshing wallpapers.

* **Tests**
  * Updated the referenced test subproject revision.


kRHYME7:
screenshots please

Hareekshith:
Now, I am able to get the refresh list menu, but I am not able to set the wallpaper, still the exact same error! 

Edit: Tried placing the wallpaper inside the Themes folder inside ./config/hyde, that worked.

Delcado19:
Good catch, and sorry — #2121 only fixed the checked-in `hydewallpaper.desktop`, but `wallpaper.kon.sh` rewrites that same file at runtime with its own bare script name, so it undid the fix on "Refresh List" and never routed the per-theme "Set As Wallpaper" clicks through `hyde-shell` at all. That's exactly the error you hit. Follow-up fix (and a regression test that would've caught it) is up in #2123 — feedback welcome there.

